PR #39, opened 2018-08-17
Added back host email setting in smtp server configuration

### LINKED ISSUE #6: Can't change host email in Dnn 9
This issue was moved here because it is only a UI issue, it was originally at dnnsoftware/Dnn.Platform#2050

In previous versions of Dnn, there was a setting in Host Settings to set the host email. This email was used when testing the SMTP server settings. So when testing the emails, users do not receive the email since they don't have access to host@change.me (the default on new installations)

Also, some modules may use that email address to notify the host, so people changing their email and not having control over this, stop receiving those notifications.

I think this could be added in Settings -> Server Settings -> SMTP Server

In the database, this is stored in HostSettings Table with a SettingName of HostEmail

Originally created by @valadas at https://github.com/dnnsoftware/Dnn.AdminExperience.Extensions/issues/604

valadas:
# Testing steps:
- Go to Settings -> Servers -> Server Settings -> SMTP Server
- Change the host email and save
- Go to Settings -> SQL Console
- Enter the following query: 
```SQL
SELECT SettingValue FROM {databaseOwner}{objectQualifier}HostSettings
WHERE SettingName = 'HostEmail'
```
- Make sure the returned email is the one entered in the UI

# Context
The host email is used in a couple places in the platform, see https://github.com/dnnsoftware/Dnn.Platform/search?q=hostemail&unscoped_q=hostemail
It can also be used by module developers to send a message to the person responsible for hosting the possibly multiple sites.
Since Dnn 9 and the persona bar came out, the only way to change this host email was to edit the database directly.

closes #6 

This is my first pull request for the persona bar, if I did anything wrong or not following best practices, please comment.

### DIFF
--- a/Extensions/Settings/Dnn.PersonaBar.Servers/Servers.Web/src/components/Tabs/SmtpServer.jsx
+++ b/Extensions/Settings/Dnn.PersonaBar.Servers/Servers.Web/src/components/Tabs/SmtpServer.jsx
@@ -61,6 +61,7 @@ class SmtpServer extends Component {
             smtpAuthentication: smtpSettings.smtpAuthentication,
             smtpUsername: smtpSettings.smtpUserName,
             smtpPassword: smtpSettings.smtpPassword,
+            smtpHostEmail: smtpSettings.smtpHostEmail,
             enableSmtpSsl: smtpSettings.enableSmtpSsl,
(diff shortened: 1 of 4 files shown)


ohine (approved):

mitchelsellers (approved):
LGTM

PR #40, opened 2018-08-20
bump version to 1.6.1.0 for Dnn v9.2.2 rc0 build

ohine:
version bump to v1.6.1.0 for generating dnn v9.2.2 rc package

### DIFF
--- a/Build/Symbols/Symbols.dnn
+++ b/Build/Symbols/Symbols.dnn
@@ -1,6 +1,6 @@
-﻿<dotnetnuke type="Package" version="5.0">
+<dotnetnuke type="Package" version="5.0">
   <packages>
-    <package name="Dnn.PersonaBar.UI_Symbols" type="Library" version="07.01.02">
+    <package name="Dnn.PersonaBar.UI_Symbols" type="Library" version="01.06.01">
       <friendlyName>DotNetNuke Dnn.PersonaBar.UI Symbols</friendlyName>
       <description>This package contains Debug Symbols and Intellisense files for DotNetNuke Dnn.PersonaBar.UI Edition.</description>
       <owner>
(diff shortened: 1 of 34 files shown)


Build/Symbols/Symbols.dnn
```diff
@@ -1,6 +1,6 @@
-﻿<dotnetnuke type="Package" version="5.0">
+<dotnetnuke type="Package" version="5.0">
   <packages>
-    <package name="Dnn.PersonaBar.UI_Symbols" type="Library" version="07.01.02">
+    <package name="Dnn.PersonaBar.UI_Symbols" type="Library" version="01.06.01">
```
mitchelsellers, line 3: This change will result in an install issue as it is making the version # lower than that of the installed one

Extensions/Build/Symbols/Symbols.dnn
```diff
@@ -1,6 +1,6 @@
-﻿<dotnetnuke type="Package" version="5.0">
+<dotnetnuke type="Package" version="5.0">
   <packages>
-    <package name="Dnn.PersonaBar.Extensions_Symbols" type="Library" version="07.01.02">
+    <package name="Dnn.PersonaBar.Extensions_Symbols" type="Library" version="01.06.01">
```
mitchelsellers, line 3: Similar comment on version number here

mitchelsellers (changes requested):
Looks like two of the packages need to follow a different version # format.  Possibly just make them 9.2.2?

Extensions/Build/Symbols/Symbols.dnn
ohine: Was the symbol package ever published? We don't have any published releases of the personabar, and the symbols don't appear to be on any previous release of platform... right? Or am I missing something? They'll be one more commit pushed to update the nuspec file version nodes in a few minutes....

mitchelsellers (approved):
Based on the conversation, the symbols have not previously been published this change isn't impacting the greater community, approved.

Future releases should look at using a different Version number format.

ohine:
Just pushed a commit with updated nuspec files. That should be everything required. I also updated the powershell script which I'm using via VSTS to set these version numbers for the next release.

mitchelsellers (approved):

PR #49, opened 2018-08-23
merge changes from release/1.6.1 back down to development

ohine:


ohine:
The referenced changes have already been approved via #47. Since this is a release management task we are self approving the merge without any reviews.

PR #54, opened 2018-08-28
Broken sorting in SQLConsole

### LINKED ISSUE #55: SQL Console - Query results - Sorting not working as expected
## Description

Sorting does apply correct order to SQL console results.

## Steps to reproduce

* Execute a select query to get results in the results section.
* In the Results Section, click on any column name to sort in ascending order.
* In the Results Section, click on same column name to sort in descending order.
* Click on any other column to sort as per ascending order of that column

## Current behavior

Sorting is not working as expected.

image-2018-08-21-13-06-38-625

## Expected behavior

Sorting should be working as expected.

mean2me:
Fixed broken javascript for table sorting in PB -> Settings -> SQLConsole

fixes #55

### DIFF
--- a/Extensions/Settings/Dnn.PersonaBar.SqlConsole/admin/personaBar/scripts/sort.js
+++ b/Extensions/Settings/Dnn.PersonaBar.SqlConsole/admin/personaBar/scripts/sort.js
@@ -24,7 +24,7 @@ define(['jquery', 'knockout'], function ($, ko) {
 
     var changeSort = function (e) {
         var $element = $(e.target);
-        var viewModel = ko.dataFor($element.parent()[0]);
+        var viewModel = ko.dataFor($element.parent().parent().parent()[0]);
         var column = ko.dataFor($element[0]);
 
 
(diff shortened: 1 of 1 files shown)


Extensions/Settings/Dnn.PersonaBar.SqlConsole/admin/personaBar/scripts/sort.js
```diff
@@ -24,7 +24,7 @@ define(['jquery', 'knockout'], function ($, ko) {
 
     var changeSort = function (e) {
         var $element = $(e.target);
-        var viewModel = ko.dataFor($element.parent()[0]);
+        var viewModel = ko.dataFor($element.parent().parent().parent()[0]);
```
tpluscode, line 27: This is an awkward construct and potentially brittle in case of changes to element hierarchy. Is there no other way to get that view model?

mean2me: If we plan to redesign this module and migrate to a modern framework, I would totally agree and we can of course make it better. I just fixed a bug avoiding any kind of breaking change. I don't think it's worth to refactor a Knockout based module since it won't be anything much better than now. If you don't like the .parent() approach, we can use DOM ids or class names, but it would be an enhancement, not a fix and it would require changes in current DOM structure.

tpluscode: Sure, that's why I marked this as comment and not "changes requested". 

Yes, changes to existing DOM structured won't do much good. I'm more asking if this is the best approach for current codebase.

tpluscode (approved):
Guess we should let it go for legacy jQuery/knockout components. Agreed @dnnsoftware/tag

mitchelsellers (approved):
The TAG group has discussed this and approving as-is.

PR #62, opened 2018-08-31
In PB -> Search Users: exact text search replaced by prefix text search

mean2me:
## Summary
In PB -> Search Users: exact text search replaced by prefix text search, since it's what any user expects.


sleupold:
it would be nice, if "*anything" would still allow a search in the middle (although with bad performance)

mitchelsellers (changes requested):
This code introduces a redundant call into the SearchTextFilter.CleanWildcards method that should not be here.  It is already called in line 513.

ohine (changes requested):
Broken unit tests need to be fixed. https://dotnetfoundation.visualstudio.com/DNN/_build/results?buildId=4502&view=ms.vss-test-web.test-result-details

mitchelsellers (approved):
Thanks for the fix

ohine:
@mean2me We still have failing unit tests related to this change. This will need to be fixed before we can merge this pull request.

tpluscode (approved):

tpluscode:
@ohine all good here?

ohine (approved):

PR #72, opened 2018-09-05
Portal Creation Overwrites Portal Folder

### LINKED ISSUE #2292: Portal Creation Overwrites Portal Folder
## Steps to reproduce

1. Login as a host user in your site. 
2. Go to `Manage > Sites`, Add new site. 
3. Give it any name, but for the Directory section use `Portals/2`, create the site. 
4. Go to the new created site, and under assets create a folder, and add an asset. 
5. Verify that these exists in the file system 
6. Create a new site, but this time do not modify the Directory, it will create it using `Portals/[Portal ID]` 
7. Visit the Assets for the first portal created and resync the file system.

## Current result

All files/folders created are gone. This happens because the Portal first created was PortalID 1, but we created it so it uses a folder with the name '2'; therefore, once PortalID 2 was created it overwrote the folder entirely, and at this point both Portals share the same folder.

## Expected result

The user should get a warning if the Portal folder already exists which should stop users from overwriting it. 

This is a niche issue; however, it can cause the loss of data, if a user uses some odd enumeration for their folders.

## Affected versions

* [x] 9.2.1

mean2me:
## Summary
Avoided creation of portals using an existing home folder.

fixes dnnsoftware/Dnn.Platform#2292

### DIFF
--- a/Extensions/Manage/Dnn.PersonaBar.Sites/Components/SitesController.cs
+++ b/Extensions/Manage/Dnn.PersonaBar.Sites/Components/SitesController.cs
@@ -425,10 +425,11 @@ namespace Dnn.PersonaBar.Sites.Components
             //Validate Home Folder
             if (!string.IsNullOrEmpty(homeDir))
             {
-                if (string.IsNullOrEmpty(string.Format("{0}\\{1}\\", Globals.ApplicationMapPath, homeDir).Replace("/", "\\")))
+                var fullHomeDir = string.Format("{0}\\{1}\\", Globals.ApplicationMapPath, homeDir).Replace("/", "\\");
+                if (Directory.Exists(fullHomeDir))
                 {
                     error = true;
-                    message = Localization.GetString("InvalidHomeFolder", LocalResourcesFile);
+                    message = string.Format(Localization.GetString("CreatePortalHomeFolderExists.Error", LocalResourcesFile), homeDir);
                 }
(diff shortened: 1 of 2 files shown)


Extensions/Manage/Dnn.PersonaBar.Sites/Components/SitesController.cs
```diff
@@ -425,10 +425,11 @@ public string ExportPortalTemplate(ExportTemplateRequest request, UserInfo userI
             //Validate Home Folder
             if (!string.IsNullOrEmpty(homeDir))
             {
-                if (string.IsNullOrEmpty(string.Format("{0}\\{1}\\", Globals.ApplicationMapPath, homeDir).Replace("/", "\\")))
+                var fullHomeDir = string.Format("{0}\\{1}\\", Globals.ApplicationMapPath, homeDir).Replace("/", "\\");
+                if (Directory.Exists(fullHomeDir))
```
tpluscode, line 429: Maybe this check should actually be done by reading HomeDirectory column from Portals table? @dnnsoftware/tag

mean2me: @tpluscode not sure it already exists at this point. Let me verify.

tpluscode:
I have to close this PR. We need those changes as a hot fix to v1.5.0. Will reopen once we have a correct branch

PR #82, opened 2018-09-19
Set default permission to "View Folder".

### LINKED ISSUE #2324: Incorrect behaviour while adding the new role to the folder permission
## Description of bug

On adding the new role to the folder of site assets then

1. Default Permission Value for newly added role should be "View Folder".
2. After saving the Newly added role. That role should persists in the application.

## Steps to reproduce

1. Login as a Superuser.
2. Go to Site Assets. 
3. Click on Edit button in the folder. 
4. Go to the permission tab. 
5. Add New Global Role. 
6. Click on Save Button. 
7. Go to Home Page. 
8. Again Search the permissions of same folder.

## Current result

1. No Default Permission Value is selected for newly added role. 
2. After saving the Newly added role. again if we check the permission then newly added role is not present in the application.

## Expected result

1. Default Permission Value for newly added role should be "View Folder". 
2. After saving the Newly added role. That role should persists in the application.

## Affected version

* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser

* [ ] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

mean2me:
## Summary
When we add a new role to *assets' permission grid* expected behavior is to have a new permissions row having *View Folder* permission set to true.

fixes dnnsoftware/Dnn.Platform#2324

### DIFF
--- a/Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
+++ b/Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
@@ -8,16 +8,6 @@ if (typeof dnn.controls === "undefined" || dnn.controls === null) { dnn.controls
 
 (function ($) {
 	
-	var permissionsMap = {
-		BROWSE_FOLDER: 8,
-		VIEW_FOLDER: 5,
-		ADD: 28,
-		COPY: 29,
-		DELETE: 30,
-		MANAGE_SETTINGS: 31,
-		WRITE_TO_FOLDER: 6
-	};
-	
     var permissionGrid = dnn.controls.PermissionGrid = function (parent, data, options) {
         this.options = options;
         this.data = data;
@@ -37,6 +27,17 @@ if (typeof dnn.controls === "undefined" || dnn.controls === null) { dnn.controls
             window.require(['css!../../css/permissionGrid.css',
                     'css!../../../../../Resources/Shared/Components/Tokeninput/Themes/token-input-facebook.css'
             ]);
+        },
+        getPermissionsDefinition: function () {
+            var permissionsDefinitions = this.data.permissionDefinitions;
+            var permissionsMap = {};
+            permissionsDefinitions.forEach(function (curr) {
+
+                var label = curr.permissionName.toUpperCase().replace(' ', '_');
+                var value = curr.permissionId;
+                permissionsMap[label] = value;
+            });
+            return permissionsMap;
         },
 
         getPermissions: function() { //get permissions from table
@@ -374,8 +375,9 @@ if (typeof dnn.controls === "undefined" || dnn.controls === null) { dnn.controls
 
             var defaultPermission = this.data.permissionDefinitions.find(
                 function(permission) {
-                    return permission.permissionId == permissionsMap.VIEW_FOLDER;
-                }
+                    return permission.permissionId == this.getPermissionsDefinition().VIEW_FOLDER;
+                },
+                this
             );
             defaultPermission.allowAccess = true;
             this._buildGridRow(this._rolesTable.find('tbody'), {
--- a/Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
+++ b/Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
@@ -361,10 +361,16 @@ if (typeof dnn.controls === "undefined" || dnn.controls === null) { dnn.controls
                 return false;
             }
 
+            var defaultPermission = this.data.permissionDefinitions.find(
+                function(permission) {
+                    return permission.permissionId == 5;
+                }
+            );
+            defaultPermission.allowAccess = true;
             this._buildGridRow(this._rolesTable.find('tbody'), {
                 roleId: roleId,
                 roleName: roleName,
-                permissions: []
+                permissions: [defaultPermission]
             }, 'roles');
 
             if (typeof this.options.onPermissionChanged == "function") {
--- a/Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
+++ b/Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
@@ -17,6 +17,16 @@ if (typeof dnn.controls === "undefined" || dnn.controls === null) { dnn.controls
         return this;
     };
 
+    var permissionLabel = {
+        BROWSE_FOLDER: 'BROWSE_FOLDER',
+        VIEW_FOLDER: 'VIEW_FOLDER',
+        ADD: 'ADD',
+        COPY: 'COPY',
+        DELETE: 'DELETE',
+        MANAGE_SETTINGS: 'MANAGE_SETTINGS',
+        WRITE_TO_FOLDER: 'WRITE_TO_FOLDER'
+    };
+
     dnn.controls.PermissionGrid.prototype = {
         constructor: permissionGrid,
         init: function () {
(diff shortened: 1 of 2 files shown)


Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
```diff
@@ -361,10 +361,16 @@ if (typeof dnn.controls === "undefined" || dnn.controls === null) { dnn.controls
                 return false;
             }
 
+            var defaultPermission = this.data.permissionDefinitions.find(
+                function(permission) {
+                    return permission.permissionId == 5;
```
tpluscode: Please change so that we don't have a magical `5` here. At least let's introduce a friendly-named constant

tpluscode (changes requested):

Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
```diff
@@ -361,10 +361,16 @@ if (typeof dnn.controls === "undefined" || dnn.controls === null) { dnn.controls
...
+                function(permission) {
+                    return permission.permissionId == 5;
+                }
+            );
+            defaultPermission.allowAccess = true;
             this._buildGridRow(this._rolesTable.find('tbody'), {
                 roleId: roleId,
                 roleName: roleName,
-                permissions: []
+                permissions: [defaultPermission]
```
tpluscode, line 379: Why is it this particular permission though?

mean2me: This is what we need to have the permission checked for "View Folder" as required.

Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
mean2me: I will introduce a permissions map.

tpluscode (approved):

tpluscode:
Please review @dnnsoftware/tag

Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
```diff
@@ -7,6 +7,17 @@ if (typeof dnn === "undefined" || dnn === null) { dnn = {}; }
 if (typeof dnn.controls === "undefined" || dnn.controls === null) { dnn.controls = {}; }
 
 (function ($) {
+	
+	var permissionsMap = {
+		BROWSE_FOLDER: 8,
+		VIEW_FOLDER: 5,
```
bdukes: Given that these are defined in the database, I'd be more comfortable loading these IDs from the server, rather than assuming they're always these hardcoded values (unless I'm misunderstanding and these are also hardcoded elsewhere)

bdukes (changes requested):

Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
```diff
@@ -7,6 +7,17 @@ if (typeof dnn === "undefined" || dnn === null) { dnn = {}; }
...
+	var permissionsMap = {
+		BROWSE_FOLDER: 8,
+		VIEW_FOLDER: 5,
+		ADD: 28,
+		COPY: 29,
+		DELETE: 30,
+		MANAGE_SETTINGS: 31,
+		WRITE_TO_FOLDER: 6
+	};
+
```
daguiler: Indentation convention in this file is using spaces, not tabs.

Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
tpluscode: @mean2me do you know how this is handled in other places in code? A `GET` endpoint to retrieve these values does make sense. Especially that you introduced the exact same thing in #83

tpluscode (changes requested):
I have to agree with @bdukes. Looking at the `*.sqldataprovider` files it's clear that permission's ids are not guaranteed to be constant. 

@mean2me clearly, permissions must already be loaded from the back end to fill in the checkbox grid. Once we get hold of them the string `PermissionCode` should be used to look up the desired value.

Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
```diff
@@ -27,7 +38,20 @@ if (typeof dnn.controls === "undefined" || dnn.controls === null) { dnn.controls
                     'css!../../../../../Resources/Shared/Components/Tokeninput/Themes/token-input-facebook.css'
             ]);
         },
-
+        _getPermissionsDefinition: function () {
+            var permissionsDefinitions = this.data.permissionDefinitions;
+            var permissionsMap = {};
+            permissionsDefinitions.forEach(function (curr) {
+
+                var label = curr.permissionName.toUpperCase().replace(' ', '_');
```
tpluscode: `data.permissionDefinitions` does not include the string identifier which we have in the database?

tpluscode (changes requested):
Much better, thank you @mean2me.

However, I think it's now unnecessarily complicated. Instead of getting the id and then looking up the permission through `this.data.permissionDefinitions.find()` call you could simply look through permissions and return the right one.

Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
mean2me: actually I think it's more readable if we use "enum" style and, it also prevent case related issues. Anybody reading this code, now, will have a clear understanding of what happens in this file, IMO.

mean2me: In our DB we have a text label containing spaces and upper/lower case letters. I feel safer if I normalize those strings.

sleupold: @mean2me you should rely on PermissionCode and PermissionKey, not PermissionName

tpluscode: > I feel safer if I normalize those strings

but you are not safer. for all I know, the names can change and the `PermissionKey` does not

mean2me: I agree, guys, it means I need to change the API since permissionKey is not returned by *GET /API/personaBar/assets/GetFolderDetails?folderId=<ID>*

Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
```diff
@@ -361,10 +366,13 @@ if (typeof dnn.controls === "undefined" || dnn.controls === null) { dnn.controls
                 return false;
             }
 
+            var defaultPermission = this.getPermission('READ');
```
tpluscode: Now I understand what @sleupold was saying. To select the correct permission you have to select by the combination of key and code. The `View folder` permission could be selected as

```suggestion
            var defaultPermission = this.getPermission({
                code: 'SYSTEM_FOLDER',
                key: 'READ'
            });
```

tpluscode (approved):

tpluscode:
I think we're good now. @daguiler @bdukes @sleupold ?

daguiler:
Yes, LGTM, let's wait for @bdukes and @sleupold.

bdukes (approved):
👍

PR #83, opened 2018-09-19
Set default permission to "View Folder".

### LINKED ISSUE #2325: Incorrect behaviour while adding the new user to the folder permission
## Description of bug
On adding new user to the folder of site assets then
1. Default Permission Value should be "View Folder" when user is added.
2. After saving the Newly added user. That user should persist in the application.

## Steps to reproduce
1. Login as a Superuser.
2. Go to Site Assets. 
3. Click on Edit button in the folder. 
4. Go to the permission tab. 
5. Type first two letters of User. 
6. Click on Add button 
7. Click on Save Button. 
8. Go to Home Page. 
9. Again Search the permissions of same folder. 

## Current result
1. No Default Permission Value is selected for newly added user. 
2. After saving the Newly added user, again if we check the permission tab then newly added user is not present in the application. 

## Expected result
1. Default Permission Value should be "View Folder" when user is added. 
2. After saving the Newly added user. That user should persist in the application.

mean2me:
## Summary
When we add a new permission role to an asset folder, expected behavior is to have a new permissions row with "view folder" permission enabled.

fixes dnnsoftware/Dnn.Platform#2325

### DIFF
--- a/Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
+++ b/Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
@@ -382,10 +382,17 @@ if (typeof dnn.controls === "undefined" || dnn.controls === null) { dnn.controls
             for (var i = 0; i < selectedUsers.length; i++) {
                 var user = selectedUsers[i];
                 if (!this._userExistsInGrid(user.id)) {
+
+                    var defaultPermission = this.data.permissionDefinitions.find(
+                        function(permission) {
+                            return permission.permissionId == 5;
+                        }
+                    );
+                    defaultPermission.allowAccess = true;
                     this._buildGridRow(this._usersTable.find('tbody'), {
                         userId: user.id,
                         displayName: user.name,
-                        permissions: []
+                        permissions: [defaultPermission]
                     }, 'users');
                 }
             }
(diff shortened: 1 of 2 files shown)


Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
```diff
@@ -382,10 +382,17 @@ if (typeof dnn.controls === "undefined" || dnn.controls === null) { dnn.controls
             for (var i = 0; i < selectedUsers.length; i++) {
                 var user = selectedUsers[i];
                 if (!this._userExistsInGrid(user.id)) {
+
+                    var defaultPermission = this.data.permissionDefinitions.find(
+                        function(permission) {
+                            return permission.permissionId == 5;
```
tpluscode: Please extract a constant for the magical number

tpluscode (commented):
This is almost identical to #82. Please think if they can be combined to avoid duplication

tpluscode (changes requested):
Please address comments

Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/permissionGrid.js
```diff
@@ -7,6 +7,17 @@ if (typeof dnn === "undefined" || dnn === null) { dnn = {}; }
...
+	
+	var permissionsMap = {
+		BROWSE_FOLDER: 8,
+		VIEW_FOLDER: 5,
+		ADD: 28,
+		COPY: 29,
+		DELETE: 30,
+		MANAGE_SETTINGS: 31,
+		WRITE_TO_FOLDER: 6
+	};
```
tpluscode: Please see comments from #82

tpluscode (changes requested):

tpluscode:
#82 just got merged. Please update the code here

PR #102, opened 2018-10-17
Fixed issues in SqlConsole sorting.

mean2me:
Fixed issues in SqlConsole sorting.

### DIFF
--- a/Extensions/Settings/Dnn.PersonaBar.SqlConsole/admin/personaBar/scripts/SqlConsole.js
+++ b/Extensions/Settings/Dnn.PersonaBar.SqlConsole/admin/personaBar/scripts/SqlConsole.js
@@ -262,7 +262,7 @@ define(['jquery',
                 case 'boolean':
                     return value;
                 case 'number':
-                    return parseFloat(value);
+                    return value;
                 case 'string':
                     var lcValue = value.toLowerCase();
                     if(isFinite(lcValue)) {
--- a/Extensions/Settings/Dnn.PersonaBar.SqlConsole/admin/personaBar/scripts/SqlConsole.js
+++ b/Extensions/Settings/Dnn.PersonaBar.SqlConsole/admin/personaBar/scripts/SqlConsole.js
@@ -9,16 +9,16 @@ All Rights Reserved
 * Module responsible to Sql Console
 */
 define(['jquery',
-        'knockout',
-        'knockout.mapping',
-        'main/codeEditor',
-        './sort',
-        './exportData',
-        './clipboard.min',
-        './html2canvas',
-        './FileSaver.min',
-        'dnn.jquery',
-        'main/koBindingHandlers/jScrollPane'],
+            'knockout',
+            'knockout.mapping',
+            'main/codeEditor',
+            './sort',
+            './exportData',
+            './clipboard.min',
+            './html2canvas',
+            './FileSaver.min',
+            'dnn.jquery',
+            'main/koBindingHandlers/jScrollPane'],
     function ($, ko, koMapping, codeEditor, sort, exportData, Clipboard) {
         'use strict';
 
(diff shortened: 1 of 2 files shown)


Extensions/Settings/Dnn.PersonaBar.SqlConsole/admin/personaBar/scripts/SqlConsole.js
```diff
@@ -355,6 +358,10 @@ define(['jquery',
                             rightData = rightData.toLowerCase();
                         }
 
+                        if(isNaN(leftData) && isNaN(rightData)) {
+                            return 0;
+                        }
```
tpluscode: Uh oh, this won't work. The compared values can be numbers, text or boolean. The latter two will always be `NaN` and thus break sorting.

Extensions/Settings/Dnn.PersonaBar.SqlConsole/admin/personaBar/scripts/SqlConsole.js
```diff
@@ -355,6 +358,10 @@ define(['jquery',
...
+
                         if (leftData === rightData) {
                             return 0;
                         } else if (leftData < rightData) {
```
tpluscode: We also have a problem here. I noticed that somewhere between getting data from the server and sorting boolean `true` value gets changed into a string `"true"`. It also breaks the comparison which happens to be like `false < "true"` and `"true" < false" instead of proper bool comparison.

tpluscode (changes requested):

Extensions/Settings/Dnn.PersonaBar.SqlConsole/admin/personaBar/scripts/sort.js
```diff
@@ -71,6 +74,7 @@ define(['jquery', 'knockout'], function ($, ko) {
     }
 
     var update = function(element, valueAccessor, allBindings, viewModel, bindingContext) {
+        console.log(element);
```
tpluscode: Please remember to remove this

tpluscode (changes requested):
Please make sure you use string equality `===` and remove the `console.log`

tpluscode (approved):

tpluscode:
we're ready here @dnnsoftware/approvers

Extensions/Settings/Dnn.PersonaBar.SqlConsole/admin/personaBar/scripts/sort.js
```diff
@@ -23,15 +23,17 @@ define(['jquery', 'knockout'], function ($, ko) {
     'use strict';
 
     var changeSort = function (e) {
-        var $element = $(e.target);
+        var $element = $(e.currentTarget);
+        var arrow = $element.find('.sort');
```
bdukes: Should this be named `$arrow`?

Extensions/Settings/Dnn.PersonaBar.SqlConsole/admin/personaBar/scripts/sort.js
```diff
@@ -51,14 +53,15 @@ define(['jquery', 'knockout'], function ($, ko) {
...
-              
+
         sortType = viewModel.sortType();
 
         viewModel.currentPage(1);
-        $element.parent().addClass('current');
-        $element.removeClass('asc desc');
+        $element.addClass('current');
+
+        arrow.removeClass('asc desc');
```
bdukes: This already happened in line 31, right?

```
$element.parent().find('.sort').removeClass('asc desc');
```

Extensions/Settings/Dnn.PersonaBar.SqlConsole/admin/personaBar/scripts/sort.js
bdukes:
```suggestion
        var $arrow = $element.find('.sort');
```

Extensions/Settings/Dnn.PersonaBar.SqlConsole/admin/personaBar/scripts/sort.js
bdukes:
```suggestion
```

Extensions/Settings/Dnn.PersonaBar.SqlConsole/admin/personaBar/scripts/SqlConsole.js
```diff
@@ -253,6 +253,32 @@ define(['jquery',
...
+        var normalizeType = function (value) {
+            if(value === "" || value === null || typeof value === "undefined") {
+                return "";
+            }
+            var type = typeof value;
+            switch(type) {
+                case 'boolean':
+                    return value;
+                case 'number':
+                    return parseFloat(value);
```
bdukes: In what scenario would calling `parseFloat` on a number return a different value than was input?

bdukes (changes requested):

mitchelsellers:
@bdukes Can you look at this one again?

bdukes (approved):
LGTM

PR #104, opened 2018-10-23
 DNN-25937: add setting/delete icon on localization pages.

### LINKED ISSUE #105: Settings/Delete localization functions is missing in page settings
## Description of bug
Settings/Delete localization functions is missing in page settings

## Steps to reproduce
1. install evoq 9x and login as host; 
2. add a new language and enable content localization in Site Settings; 
3. try to edit any page and switch to Localization tab.

## Current result
only have the view icon, but no gear icon or trash icon.

## Expected result
same functionality with 8x: 
eye icon: click to browse the page; 
gear icon: click to check the page setting of that localized page; 
trash icon: click to delete that localized culture of the page.

## Screenshots
image
image

* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [x] 9.1
* [x] 9.0

zyhfish:
Fix #105.


Extensions/Content/Dnn.PersonaBar.Pages/Components/IPagesController.cs
```diff
@@ -23,7 +23,7 @@ public interface IPagesController
 
         TabInfo MovePage(PageMoveRequest request);
 
-        void DeletePage(PageItem page, PortalSettings portalSettings = null);
+        void DeletePage(PageItem page, bool hardDelete = false, PortalSettings portalSettings = null);
```
tpluscode: Adding parameter to an interface method is a breaking change. @mitchelsellers ?

Extensions/Content/Dnn.PersonaBar.Pages/admin/personaBar/App_LocalResources/Pages.resx
```diff
@@ -1690,4 +1690,7 @@
   <data name="EditContent.Text" xml:space="preserve">
     <value>Edit Content</value>
   </data>
+  <data name="DeleteTranslationConfirm.Text" xml:space="preserve">
+    <value>You are about to permanently &lt;b&gt;emove&lt;/b&gt; the '&lt;b&gt;{0}&lt;/b&gt;' translation of this page.  Are you sure you want to proceed?</value>
```
tpluscode:
```suggestion
    <value>You are about to permanently &lt;b&gt;remove&lt;/b&gt; the '&lt;b&gt;{0}&lt;/b&gt;' translation of this page. Are you sure you want to proceed?</value>
```

Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/services/pageService.js
```diff
@@ -23,9 +23,12 @@ const PageService = function () {
         return api.post("SavePageDetails", toBackEndPage(request));
     };
 
-    const deletePage = function (page) {
+    const deletePage = function (page, hardDelete) {
         const api = getPagesApi();
-        return api.post("DeletePage", { id: page.tabId });
+        if(typeof hardDelete === "undefined"){
+            hardDelete = false;
+        }
```
tpluscode: We could use default parameter values

```js
const deletePage = function (page, hardDelete = false) {
}
```

Extensions/Content/Dnn.PersonaBar.Pages/Components/Prompt/Commands/DeletePage.cs
```diff
@@ -55,7 +55,7 @@ public override ConsoleResultModel Run()
             }
             try
             {
-                PagesController.Instance.DeletePage(new PageItem { Id = PageId }, PortalSettings);
+                PagesController.Instance.DeletePage(new PageItem { Id = PageId }, false, PortalSettings);
```
tpluscode: By the way why add the new parameter in the middle? If it was added to the end of the signature then this usage would not have to have been touched.

Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/actions/pageActions.js
```diff
@@ -206,26 +206,30 @@ const pageActions = {
...
                     return;
                 }
 
-                dispatch({
-                    type: ActionTypes.DELETED_PAGE
-                });
-                if (page.tabId !== 0 && (page.tabId === utils.getCurrentPageId()) || redirectUrl) {
-                    window.top.location.href = redirectUrl ? redirectUrl : utils.getDefaultPageUrl();
+                if(typeof callback === "function"){
+                    callback();
```
tpluscode: Using `Promise` or even `async/await` would be better than adding a callback parameter

Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageLocalization/PageLocalization.jsx
```diff
@@ -39,6 +41,27 @@ class PageLocalization extends Component {
...
+        }
+    }
+
+    onDeletePage(page){
+        let callback = () => {
+            this.removeLocaleFromState(page.CultureCode);
+            if(page.TabId === utils.getCurrentPageId()){
+                let panelId = window.$('.socialpanel:visible').attr('id');
+                utils.getUtilities().panelViewData(panelId, {tab: [0]});
+                window.top.location.href = utils.getDefaultPageUrl();
```
tpluscode, line 62: is `window.location` the default choice to navigate the browser between pages?

tpluscode (changes requested):
Does the term *hard delete* appear elsewhere in DNN? I would tentatively suggest to call it *permanent* deletion.

Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/services/pageService.js
zyhfish: hard delete used in TabController as well, so i keep the same name here.

tpluscode: 👍

Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/actions/pageActions.js
zyhfish: because here will prevent the different behavior after page deleted, and add custom behavior, so add callback to avoid redirectUrl stuff.

Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageLocalization/PageLocalization.jsx
zyhfish: Yes

Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/actions/pageActions.js
tpluscode: In which case maybe this should be another `dispatch` instead?

Extensions/Content/Dnn.PersonaBar.Pages/Services/PagesController.cs
```diff
@@ -303,7 +303,7 @@ public HttpResponseMessage DeletePage(PageItem page)
 
             try
             {
-                _pagesController.DeletePage(page);
+                _pagesController.DeletePage(page, PortalSettings, hardDelete);
```
tpluscode: hm, the code before was not passing in `PortalSettings`. will it be the same object?

zyhfish: Yes, they both call to current portal settings.

Extensions/Content/Dnn.PersonaBar.Pages/Components/IPagesController.cs
mitchelsellers: @tpluscode Yes, this would be a breaking change for any consumer of the interface that was compiled.

tpluscode: @zyhfish please make the new method an overload

Extensions/Content/Dnn.PersonaBar.Pages/Components/IPagesController.cs
```diff
@@ -23,7 +23,7 @@ public interface IPagesController
 
         TabInfo MovePage(PageMoveRequest request);
 
-        void DeletePage(PageItem page, PortalSettings portalSettings = null);
+        void DeletePage(PageItem page, PortalSettings portalSettings = null, bool hardDelete = false);
```
tpluscode: @zyhfish please make the new method an overload

tpluscode (dismissed):
@mitchelsellers @bdukes to you think that other than the interface method we're good here?

mitchelsellers:
@tpluscode Yes, I believe if we make that an overload it no longer breaks the interface structure so it should be good.

Extensions/Content/Dnn.PersonaBar.Pages/Components/IPagesController.cs
zyhfish: updated.

daguiler (approved):

daguiler:
@tpluscode, @mitchelsellers, requested changes were implemented

daguiler:
This one's been sitting here for too long. Can you please revisit @mitchelsellers?

mitchelsellers (approved):

Issue #106, opened 2018-10-23
Persona Bar: Users: Role Names overflowing into next line


noor-zubair:
Description of bug
When a Role has a long name such as "Archives content manager Archives content manage 1" it appears to be overlapped in Roles dropdown when assigning role to a User

Steps to reproduce
1. Go to Manage > Roles 
2. Create a couple of roles with a long name such as "Archives content manager Archives content manage 1" and "Archives content manager Archives content manage 2" 
3. Go to Manage > Users 
4. Click on the shield icon for one of the users to add a role to it 
5. Type an A in the search for the roles

Current result
The text jumps the next line causing text to look overlapped. See screenshot
![screenshot-1](https://user-images.githubusercontent.com/38560146/47369843-f75df500-d6fd-11e8-9c50-df126becd3fb.png)

Expected result
It should display the name without overlapping

Affected version
  9.2.1

valadas:
Manually closing since #107 has fixed this.

PR #107, opened 2018-10-23
DNN-25490 Persona Bar: Users: Role Names overflowing into next line

noor-zubair:
Fix delivers to issue [#106](https://github.com/dnnsoftware/Dnn.AdminExperience/issues/106)

### DIFF
--- a/Extensions/Manage/Dnn.PersonaBar.Users/Users.Web/src/_exportables/src/components/UserTable/UsersRoles/index.jsx
+++ b/Extensions/Manage/Dnn.PersonaBar.Users/Users.Web/src/_exportables/src/components/UserTable/UsersRoles/index.jsx
@@ -187,7 +187,7 @@ class UserRoles extends Component {
             <div className="header">
                 <div className="header-title">{Localization.get("Roles.Title") }</div>
                 <div className="add-box">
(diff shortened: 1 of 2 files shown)


tpluscode (changes requested):
I don't think it's enough to just resize columns. Long enough text would still overflow, right?

To make it more future-proof, I'd rather see a CSS [`text-overflow: ellipsis`](https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/) maybe?

noor-zubair:
Role name field has max length defined which is 50 and the example role name used is of max length. So it will not overflow.
db

sleupold:
@noor-zubair, @tpluscode 
there is already a feature request to increase length of this field (which makes sense for roles synched from AD), i.e. I agree to the suggestion from @tpluscode.

noor-zubair:
@tpluscode ,
I am implementing the suggestion, thanks.

mikebigun:
@tpluscode @sleupold 

See text-overflow in action:

image

tpluscode (approved):
@dnnsoftware/approvers, looking good? :)

bdukes (approved):

sleupold:
looks fine for me

Issue #33, opened 2018-08-16
Make the PersonaBar navigable by keyboard 


OllyHodgson:
## Description of problem

Part of #21. The PersonaBar is not keyboard-navigable. Only users with a mouse can administer a DNN 9.x website.

## Description of solution

Users should be able to use the keyboard to navigate through the items in the PersonaBar. At the very least, they should be able to use the Tab key to focus elements, and the Enter key to activate them. 

By "activate" I mean show the relevant menu *or* in the case of a menu item / edit button take the approriate action.

## Additional context

[main.js](https://github.com/dnnsoftware/Dnn.AdminExperience/blob/development/Library/Dnn.PersonaBar.UI/admin/personaBar/scripts/main.js) handles the hover events currently used to show and hide the menus. It could be refactored to allow for menus to be shown and hidden when the relevant items are activated (or recieve focus).

Right now, the JS events to show/hide the menus are bound to `<li>` elements which cannot be focussed without additional work (at the very least we'll need `role='button'` and `tabindex='0'`). It would make sense to to use `<a>` or `<button>` elements instead - they're the right tool for the job, and can be focussed "for free". That could be broken out into a separate issue though.

## Affected version

* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [x] 9.1
* [x] 9.0

## Affected browser

All browsers.

OllyHodgson:
I'm trying to figure out a good pattern for keyboard navigation through the PersonaBar's top level menus. I'm thinking:

- `[tab]` to focus items in the menu
- `[enter]` to open sub-menus / activate the link
- `[esc]` to close the currently opened menu

Does this sound sensible? @Cambridgeport90 do you have any input here?

Cambridgeport90:
That's about the way things are normally done, though if you ask me, this should have been built into the process from the beginning. Normally this oughtn't take this much work. Something seems very off here, to be honest. But tabbing through dialogues is normally how it works, though arrows should bring you from one element to another as well. Comboboxes should be able to be opened with a single press of alt+ down arrow, and then the arrows should bring you from one item to another, enter to select an item. Tab can bring you between controls as well, though screen readers have their own sets of control keys as well, so ensure that they can still be used too; E for edit  box, F for form field, H for Heading, I for list, N for next text area, and so on.

OllyHodgson:
Thanks @Cambridgeport90, that's super helpful. When I get time, I'll get started!

[...]

PR #110, opened 2018-10-26
First go at keyboard navigation in the PersonaBar

OllyHodgson:
First shot at fixing #33.

## Summary

With this pull request:

- Buttons in the left hand menu should be keyboard focusable
- Mouse navigation works the same as before
- When a button is focussed, pressing up or down keys will focus the previous/next button. If it has a hover menu, pressing right or enter will open the hover menu and pass focus to the first menu item. If it does not have a menu, pressing enter will trigger the button's action, e.g. edit the page
- When in a hover menu, up and down keys will focus the previous/next menu item. Left or Escape will close the hover menu. Enter will open the item.

I was still having some issues with NVDA, so this probably isn't ready to merge just yet. I'd appreciate thorough testing and comments though - especially from users of screen readers and other assistive tech! For instance the current keybinding of up/down makes sense to a sighted user (because it's a vertical menu) but I suspect normally you might use left/right to navigate menu items, and up/down to move within them (like a typical Windows menu).


OllyHodgson:
Note for ESW - there are some changes in the CSS to add `:focus` styles. I know Evoq is styled differently so you might have to make the corresponding in the Evoq PersonaBar CSS.

OllyHodgson:
Having cleared my browser cache, NVDA seems to like this patch a bit more.

valadas:
Gave it a try and looks good to me from the perspective of a guy that just likes to use a keyboard instead of a mouse. That being said, this only affects the menu and not the modules in the persona bar right? Is there an idea to give focus to the first control of the module after entering a menu or would that need to be on a module by module basis?

OllyHodgson:
@valadas This PR is only to over the left-hand menu. Passing focus is a _great_ idea, but it's not something I've looked into. Can you create another issue to cover it?

valadas:
@OllyHodgson I think you would be better at documenting this, I am just a guy that finds it nice to get to things by keyboard instead of mouse, you have far more knowledge about how to describe it right for non-visual navigation. Right now, I am not even sure if this can be done at a single place or if each module needs to set it's first control focused or with a tabindex=0 or what, I am not even sure what happens if two controls have tabindex=0...

OllyHodgson:
@valadas Raised #113 to cover this. FYI, setting `tabindex="-1"` means an element can be focussed using the tab key as per normal. `tabindex="0"` means it can only be focussed programatically (i.e. via JS). `<input autofocus />` would give an element focus, that would need to be implemented on a case by case basis.

valadas:
@OllyHodgson I just found an issue though, after being in the persona bar by keyboard, how do we get back focus on the page?

OllyHodgson:
@valadas Oh, good catch! I'll have a look.

valadas:
@OllyHodgson I know I read some documentation about marking the main content so that non visual users can easily skip the navigation and headers and get to the main content, but I saw no mention on how this was done, is there some standard keyboard shortcut that is well known to bring the focus to the main page, that may be a solution?

OllyHodgson:
@valadas As far as I can tell, my bug is because when the menu has focus, my script is swallowing all the keypress events. I need to rework it a bit so the tab key continues to work as normal.

I've seen sites where they use `<div role="main" id="maincontent">` or `<main id="maincontent">` and an vsually hidden anchor `<a href="#maincontent">`  to jump straight to the main body. But that wouldn't be much help here :man_shrugging:

valadas:
@OllyHodgson but when there is a main content how does a screen reader react to that, is there a special well known shortcut to get to the main content? What is the purpose of tagging that main content?

valadas:
Ok, I did a second round of testing on this and found a couple more issues:

1. The edit bar does not open
2. Once a module is open, keyboard navigation on the persona bar is lost.
3. The keyboard is captured which breaks other shortcuts like F5 to refresh the page.
4. aria-expanded is set to true when hovering an item but is never set back to false when leaving it.

@OllyHodgson I have read a bit on accessibility and one of the best practices is to never have generic elements (div, li, span) generate an action. Elements that cause an action should be buttons or links so that the browser inherently knows what to do with the keyboard. Do you think it would be an easier solution to make the PB menu items be buttons and adding a skiplink to the underlying page instead of managing all this using javascript ?

OllyHodgson:
Sorry for the lack of activity, I've been busy relocating from the UK to BC, Canada! 

I was tempted to make the menu items into `<button>` elements, but that would have had significant impact on styling for Platform and Evoq, so I was going to pick that up in a second PR. The only thing it would give us are native focus events - we'd still need to handle the interaction through JS because this whole menu is JS code. For now I went with the `role` and `tabindex` attributes to achieve the same goal.

The real issues here are:

* I'm trying to make a hover menu keyboard-accessible. If each menu was shown with a click things would likely be a lot simpler! Unfortunately clicking a menu icon opens the first item in the sub-menu - which doubly complicates things because it's not what we want for keyboard nav. Deep down I want to *completely* refactor the interaction on this thing but that would be a major breaking change, and a fair bit of work...
* When the user is keyboard-navigating through the PB menu items, we need to capture that and show / hide / activate the relevant thing. But we only want to capture the keyboard in that specific instance, and that's quite tricky to do. Still working on that bit.
* When the user gets to the end of our menu, we need a way to pass control back to the browser. The PB lives in an iframe, which complicates things a bit. I think a skip-link (to a known ID in the main document) might be a good way to do that.

If nobody else picks it up in the meantime, hopefully I'll get another chance to work on this in the not too distant future.

OllyHodgson:
@valadas 

> when there is a main content how does a screen reader react to that

The ones I've used show the user landmark roles on the page, e.g. Main (`<main>` or `<div role="main">`), Navigation (`<nav>` or `<div role="navigation">`, etc. The user can jump directly to those landmarks.

> is there a special well known shortcut to get to the main content?

Nothing universal that I'm aware of.

> What is the purpose of tagging that main content?

Apart from the above, one example is the "reader mode" in some browsers and extensions - they use it as one way to figure out what's the main content in a page.

valadas:
Welcome to Canada heh @OllyHodgson !
Ok, I would like to work with you on this after the current RC is out.

I don't think that switching to buttons would be a big issue, most styles use classes so we may just need to add some css to remove the default button styles and Evoq should inherit that. It would remove some of the required javascript and be more semantic.

As for clicking opening the first page of the submenu, I think this is just totally wrong, clicking with the mouse or activating with the keyboard or hovering should display the submenu, then we can just add some javascript to pass the focus to the submenu if activated by the keyboard and we are golden. In the submenu, I think we can just pass back to focus to the parent at the end of the list.

I guess we can do the same for the modules, give it focus on activating the menu link and giving back focus to the PB when there are no more controls to focus on, or we could capture Escape in the module container to close the pane and give back the focus at the start of the Persona Bar. 

I think we can easily add a skiplink (not visible until focused, so would mainly apply for keyboard navigation) between the dnn logo and the first menu, that would be a simple link to the parent frame #body with an aria-label or text such as "skip to site".

OllyHodgson:
Thanks @valadas 😁

Had a bit more of a think about this. At one point I had a version of the code which only did `event.preventDefault()` for the relevant keys (up, down, left, right, tab, enter, escape). This worked much better! The remaining issue was that the tab key was captured in the PB iframe and I couldn’t figure out a good way to pass focus to the next item in the “natural tab order” when that was outside the iframe.

I agree, changing a click to open the menu instead of the first menu item makes more sense.

valadas:
@OllyHodgson This PR is now a little out of date with current development branch. I love most of what was done but there are some small issues and now that it is out of date, probably would be easier to redo it. Would you like to get together and redo it together. If so, book me some time at https://doodle.com/valadas and we can put our head together...

ohine:
@valadas I just merged in the latest changes from development. What are the issues that are left with this? Is there anyway we can push any fixes here and get this merged before code-freeze next week? 

cc: @OllyHodgson

OllyHodgson:
There's a few things outstanding from what I remember:

1. This presently traps keyboard focus in the PB iframe. If we run out of things to tab through in the PB iframe, we need to pass focus to the next focusable item in the main document's natural tab order. We need a way to find out what that thing is. Is there an element in the main document immediately after the PB iframe, which *always* has a consistent `ID` attribute, regardless of theme or other mods? We could pass focus back to that I guess?
2. Right now it swallows all keypresses if focus is on the PB - I had a fixed version at one point, but clearly never pushed it and no longer have that computer 😞 
3. It doesn't work for the edit button.
4. Currently when a user clicks on a top-level PB item, it activates the first item in the submenu. We want to change this to match how it'll work with keyboard/hover - i.e. it shows the submenu.

Unfortunately I'm not going to have time to play much with in the next few weeks. I'm happy to help out from the sidelines if someone wants to take over from me.

valadas:
@OllyHodgson did a good summary of the issues. I don't think I will have much time either until code freeze to work on this and will try to keep some hours for fixes/testing. I think we should keep this open for reference but rething a little bit so have a more semantic persona bar.

Right now it is divs everywhere and javascript to handle all interactions. Ideally everything that produces an action should be a button and everything that produces a change in navigation should be a link, buttons and links can be styled just as divs but they automatically provide 90% of the functionality we need except hover. 
I am not a big expert in this field, but I read a little bit about this and I think that makes sense, we would just need to convert each main menu item to a button, hook a click an hover event that opens the menu and give focus to the first button in that panel, then we can attach a tab listener on the last button of the submenu that closes that and gives back focus the the button that opened it. On the module panel level we can just onload give focus the the first focusable element, make the close button be last (need to think about how to do this) and voilà we have keyboard access and semanting persona bar on all the PB. Then inside the modules 90% will be updating the common components for accessibility and then checking what else is left to do on a module by module basis (some do not use the common components)..

For mouse users the only change is that clicking the big main icon would no longer open the first module but open the list of modules (submenu) and for accessibility, with buttons you gain both labelling and keyboard navigation without complex javascript. We could have a singe js method with a data-id to know which menu each button belongs to or something along those lines.

That would be my general idea anyway, happy to discuss it, but afraid it wont happen for 9.4

valadas:
Because this PR is out of sync by a lot now and is not ready to merge, I am going to close it. I will shortly open a new one referencing this one for reference so we can go at it again.

PR #116, opened 2018-11-03
Exposing AllTabs attribute

mean2me:
When a module is shared across pages, it's not listed in *Page Settings -> Modules*. 
This change will make shared module visible and editable in the modules list of all pages.


daguiler (approved):

mitchelsellers (approved):
Great change!

PR #120, opened 2018-11-06
SEO URL Regex: Settings are saved in HostSettings but pulled out from PortalSettings

### LINKED ISSUE #119: SEO URL Regex: Settings are saved in HostSettings but pulled out from PortalSettings
## Description of bug
Application fetches SEO regular expression values (`Settings -> SEO -> Expressions`) from PortalSettings, but is always saving them to HostSettings. Due to mentioned inconsistency, users are always confusing when modifying SEO expressions. Moreover, making changes for portal N will override settings of all existing portals since it always writes to HostSettings.

## Steps to reproduce

1. Go to `Settings > SEO > URL > Regex`, and add something to the Keep in Querystring Regular Expression section. 
2. Save it and reload the page. 
(Check) The settings are saved and what was added is there. 

3. Add a Portal Specific setting for this section for the current Portal. We can run the following query to add it to Portal 0. 
```sql
  INSERT INTO PortalSettings 
  VALUES (
    0, 
    'AUM_KeepInQueryStringRegex', 
    '/nomo/\d+|/runningDefault/[^/]+|/popup/(?:true|false)|/(?:page|category|sort|tags)/[^/]+|tou/[^/]+|(/utm[^/]+/[^/]+)+', 
    1, 
    GETDATE(), 
    1, 
    GETDATE(), 
    NULL, 
    0) 
```
4. Go to the `Settings > Servers > Clear Cache` to load the new settings. 
5. Go back to `SEO > URL > Regex `
It will show the default settings we added to PortalSettings instead of what was previously saved in Host Settings. 

## Expected result
Need to make sure settings are saved on a portal level. HostsSettings should be used as a fallback when portal is not defined.

* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [x] 9.1
* [ ] 9.0

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.AdminExperience/issues/119

### DIFF
--- a/Extensions/Settings/Dnn.PersonaBar.Seo/Services/SeoController.cs
+++ b/Extensions/Settings/Dnn.PersonaBar.Seo/Services/SeoController.cs
@@ -20,6 +20,7 @@
 #endregion
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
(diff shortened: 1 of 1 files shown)


daguiler (approved):

mitchelsellers (approved):
Looks good to me.

PR #135, opened 2018-11-08
Persona Bar: Pages: Wrong Container Selected

### LINKED ISSUE #134: Persona Bar Pages: Wrong Container Selected
## Description of bug
In some cases, having multiple themes installed, on a `Pages -> [Select Page] -> Advanced -> Appearance` wrong page container is auto-selected. Instead of default one, first container in the list is highlighted. 

## Steps to reproduce
1. Go to `Manage > Themes` and select the notitle_black to be the default container.
2. Go to the file system `/Portals/_default/Skins` and create 4 copies of the Cavalier Skin.
3. Go to the `Home Page` settings and change the Container to something not default such as `notitle_white`. 
4. Refresh the page and go back to the `Home Page` settings, the new selected Container is selected. 
5. Go to the Home Page settings and change the Container to the default one `notitle_black`.
6. Refresh the page and go back to the `Home Page` settings. The `notitle` container will be selected instead of the default one since it simply selects the first one in the list. 

_Please note, this issue is not reproducible every time. Above scenario with a copying skins manually just makes us possible to make this bug visible and debug it. But customer reported this also happens with a multiple themes installed using official procedure via UI_. 

## Expected result
It should show the default container as selected instead of the first one in the list. This is confusing, and will overwrite the container settings if any page setting is saved.

## Affected version

* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.AdminExperience/issues/134

When loading appearance tab, there are two api calls:
- `onRetrieveThemes`: get list of available themes and default portal theme/container
- `onRetrieveThemeFiles`: get list of available layouts and containers

Both api's are invoked asynchronously. 

Increasing number of themes in a Skin folder makes the first call completed later than the second one. At the moment when containers list is already received we don't have default value of container. It causes application to choose the first one in the list. This also explains why this issue is not reproducible having small amount of themes. Operation of reading from file system is heavy and ideally, installing themes in a correct way also installs more containers and layouts. So it makes some balance when executing above calls. In our particular case we increased skins only and this problem became visible.

To fix it, we need to make sure above calls are executed consistently, when first is completed and all themes including default values received, we raise second one. We can use `Promise` feature and invoke second api call when asynchronous call is resolved.

### DIFF
--- a/Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/actions/themeActions.js
+++ b/Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/actions/themeActions.js
@@ -3,7 +3,7 @@ import ThemeService from "../services/themeService";
 
 const themeActions = {
     retrieveThemes() {
-        return (dispatch) => {
+        return (dispatch) => new Promise(function(callback) {
             dispatch({
                 type: ActionTypes.RETRIEVING_THEMES
             });    
@@ -18,14 +18,16 @@ const themeActions = {
                         defaultPortalLayout: response.defaultPortalLayout,
                         defaultPortalContainer: response.defaultPortalContainer
                     }
-                });  
+                });
+                callback({success: true});
             }).catch((error) => {
                 dispatch({
                     type: ActionTypes.ERROR_RETRIEVING_THEMES,
                     data: {error}
                 });
+                callback({success: false});
             });     
-        };
+        });
     },
 
     retrieveThemeFiles(themeName, level) {
--- a/Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/Appearance/Appearance.jsx
+++ b/Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/Appearance/Appearance.jsx
@@ -15,15 +15,19 @@ import style from "./style.less";
 class Appearance extends Component {
 
     componentWillMount() {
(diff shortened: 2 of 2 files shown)


ohine:
Can we combine the two calls into one? I’d much rather we have one optimized call instead of waiting for one to finish before making the second.

mikebigun:
@ohine 
We can implement that for sure, but I'm afraid this will have some negative impact on performance.
As you know, all themes/layouts/containers are reading from the file system. Those are very heavy operations. 
First call to get the themes is executed on initialization, the second call is executed every time we change page theme on UI. So selecting theme, we only want to bind layouts and containers to the selected theme. Nothing else. 
Combining it to one, we always will be reading all themes again and again. Do you think it make sense?

mikebigun:
The second reason is that it is hard to identify all consumers of those endpoints. Apart of appearance tab, who else need it? So due to this I tried to avoid changing api endpoints and their responses.

ohine:
That’s understandable, I misunderstood the exact usage. 

I’m ok with leaving it as two requests, however I think we should look into improving this in the future. We should be able to go look at the installed themes in the database instead of searching the file system. 

It’s very old legacy code that was never replaced after making a skin an actual extension type.

tpluscode (approved):
@ohine do you approve this PR then?

ohine:
@tpluscode yup, merging it now.

ohine (approved):

PR #152, opened 2018-11-19
Content Localization: shared module un-published after changes

### LINKED ISSUE #151: Content Localization: shared module un-published after changes
## Description of bug
When multiple languages installed and all pages are translated and published, shared module is disappearing from all places when changing `IsLocalized` module property on a particular page settings.

## Steps to reproduce
1. Add a module to a page, and go to `Settings > Add Module` to all pages. Save. 
2. Go to `Settings > Site Settings > Languages`, and add a second language, then enable content localization and localize all of the default pages. 
3. Visit the page with another language, and change the content of the module that is set to be added to every page. 
4. Make sure the content has changed on all of the pages of that language. 
5. Go to the page settings for one of the other pages in that language. 
6. Under Localization click on the chain next to the module added to every page. This will make it a reference instead of a copy. Then hit Update Localization. 
7. Click the chain again to make it a copy, hit update localization
8. Refresh the page and check other pages of this language

See reproduction demo: [localization.mp4](https://drive.google.com/file/d/1eehFMKI1rVPEz8cMgBaxjVFfw8AThgZW/view?usp=sharing)

## Current result
The module added to all pages is no longer visible. If you go into edit mode it is still present on all pages, but not visible.

## Expected result
This case should be handled better. It should either just break the module on the current page, or warn the user if its a module shared everywhere.

## Affected version
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.AdminExperience/issues/151

Clicking on an `Update localization`, application removes all modules for which `IsLocalized` property has been changed. Instead, system starts using reference to an original version of a module. Shared module is not getting re-published automatically, so it gets invisible in a view mode. As per customers, this is expected behavior as content of referenced module can be different.
Now users are confused when particular page settings are modified, but changes are applied to all other pages. This is because of shared module.
Customer don't want to have an effect globally and localization of a shared module property should be handled better, ideally, on a module level settings and not on a page settings. 
So, request is to add tooltip info whether module is shared or not and gray out (disable) the chain icon for a shared modules.

### DIFF
--- a/Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageLocalization/Module.jsx
+++ b/Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageLocalization/Module.jsx
@@ -35,7 +35,7 @@ class Module extends Component {
 
         if (module.Exist) {
             const className = "module-row" + (module.IsDeleted ? " deleted" : "");
-            const toolTip = ["Shared: " + module.IsShared, module.ModuleInfoHelp].join("<br>");
+            const toolTip = [Localization.get("Shared") + ": " + module.IsShared, module.ModuleInfoHelp].join("<br>");
             return <div className={className}>
                 <Tooltip
                     messages={[toolTip]}
--- a/Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageLocalization/Module.jsx
+++ b/Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageLocalization/Module.jsx
@@ -35,7 +35,7 @@ class Module extends Component {
 
         if (module.Exist) {
             const className = "module-row" + (module.IsDeleted ? " deleted" : "");
-            const toolTip = [Localization.get("Shared") + ": " + module.IsShared, module.ModuleInfoHelp].join("<br>");
+            const toolTip = [module.IsShared ? Localization.get("SharedModule") : Localization.get("IsNotSharedModule"), module.ModuleInfoHelp].join("<br>");
             return <div className={className}>
                 <Tooltip
                     messages={[toolTip]}
--- a/Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageLocalization/Module.jsx
+++ b/Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageLocalization/Module.jsx
@@ -35,9 +35,10 @@ class Module extends Component {
 
         if (module.Exist) {
             const className = "module-row" + (module.IsDeleted ? " deleted" : "");
+            const toolTip = ["Shared: " + module.IsShared, module.ModuleInfoHelp].join("<br>");
             return <div className={className}>
                 <Tooltip
-                    messages={[module.ModuleInfoHelp]}
+                    messages={[toolTip]}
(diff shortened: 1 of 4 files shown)


Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageLocalization/Module.jsx
```diff
@@ -35,9 +35,10 @@ class Module extends Component {
 
         if (module.Exist) {
             const className = "module-row" + (module.IsDeleted ? " deleted" : "");
+            const toolTip = ["Shared: " + module.IsShared, module.ModuleInfoHelp].join("<br>");
```
daguiler: @mikebigun Shouldn't this string be localized?

mikebigun: @daguiler you are right. It's done.

daguiler: Sorry to insist @mikebigun, but localization should also be applied to `module.IsShared`

mikebigun: @daguiler pls, check now.

daguiler (approved):

Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageLocalization/Module.jsx
```diff
@@ -47,7 +48,7 @@ class Module extends Component {
                 </div>}
                 {!module.IsDeleted && !isDefault && <div className="icons-container">
                     <span
-                        className={`icon float-left ${(module.IsLocalized ? " blue" : "")}`}
+                        className={`icon float-left ${(module.IsShared ? " disabled" : (module.IsLocalized ? " blue" : ""))}`}
```
bdukes, line 51: If a module is both shared and localized, we _don't_ want the `blue` class, only `disabled`, is that right?

bdukes (approved):

Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/src/components/PageLocalization/Module.jsx
mikebigun: yes, correct. Blue class just highlights the current state of a localization button/link. By default this link is a blue when module is not localized, and brightly blue when it is localized (blue class makes it brightly blue).
But in case of shared module we just want to disable (grayed out) it completely. So if we still want to provide information whether it is localized or not it is hard to do that. If button will not be grayed out but we just unbind all events, so users will confuse why it is clickable but nothing is happening. 
Exactly for this purposes we added a `Shared` label, so ideally users should go to shared module settings and proceed with required actions on a module level.

mikebigun: The point is it is hard to handle `disabled + blue` and `disabled + brightly blue`. 
So the decision was to skip blue indicator and disable it completely.

PR #156, opened 2018-11-20
Fixes the Pages Module after the dependencies updates

valadas:
## Summary
Since greenkeeper updated many things including most dependencies and webpack, the Pages module was not able to build
This pull request fixes many of those issues including:
- webpack.config.js changes
- babelrc changes
- code changes (though still using some deprecated methods)
- Updated to use the new common components structure

## What is left to do
- I tested the module with the new common components and everything appears to be working
- However there is still one use of an UNSAFE_ (deprecated) method which I could not replace without causing and endless loop. It works fine but we will need to get that sorted out before React 17, for now though, it works.

closes #149

PR #171, opened 2018-11-26
Persona Bar: Pages: Wrong Container Selected

mikebigun:
This is addition to the main PR https://github.com/dnnsoftware/Dnn.AdminExperience/pull/135
The changes done in a previous PR works fine, still they are not visible on the target environment. The reason is that changes were not compiled to the resulting js file. In this PR I have re-build react project and newly created js file is updated with above changes.

Please, note, the file was compiled using version 1.13.0 of webpack prior to that one introduced by greenkeeper. I moved to a couple weeks back to be able to compile this file using old version. We should be safe to include my changes here.


daguiler (approved):

bdukes:
@ohine @mtrutledge @nvisionative @valadas Did I see something about removing the compiled bundles from the repo as part of y'all's updates?  This PR would at least be superseded by #156, right?

ohine:
I mentioned it on slack last week, and again today. I'd like to remove all generated bundles to ensure that the CI builds are always up to date. We've had a few instances over the last few months where everyone missed the fact that the bundles included with internal builds were not up to date.

If we remove all the bundles, it will be extremely obvious when a build is broken.

daguiler:
Ok, so I guess we can close this one then.

PR #178, opened 2018-11-28
Resolve GreenKeeper Change issues for Users module

mean2me:
## Summary
After Greenkeeper's changes Users module's build was corrupted. This commit fixes and updates everything needed to make ti build again.


mitchelsellers:
@mean2me Are there any other changes included in this, or is it just build improvements?

@mtrutledge or @valadas can you review this based on the other work?

mtrutledge (changes requested):
Great work here. I just have a few changes I would like to request.

In file `/Extensions/Manage/Dnn.PersonaBar.Users/Users.Web/src/_exportables/src/components/UserTable/DetailRow/index.jsx` rootElement is never created through either a arrow function or the constructor. Please add `this.rootElement = React.createRef();` in the constructor.

Can you also please update the components to the latest @dnnsoftware/dnn-react-common@^2.0.6?  Instead of class/function imports the components are now using named imports from a single package. For example instead of 
```
import DropDown from "dnn-dropdown";
import GridCell from "dnn-gridcell";
``` 
This would change to:
```
import {Dropdown, GridCell} from "@dnnsoftware/dnn-react-common";
```
Note that `DropDown` changed to `Dropdown`. There are a couple other changes to the names of components as well:
>Collapse -> Collapsible or CollapsibleRow
DropDown -> Dropdown
SvgIcons -> You must prefix the icon name with SvgIcons. E.g. ArrowBack -> SvgIcons.ArrowBack
Tabs -> DnnTabs
OverflowWrapper -> TextOverflowWrapper

The dnn-webpack-externals also changes to `@dnnsoftware/dnn-react-common/WebpackExternals`

There were a few bug fixes that were made in dnn-react-common 2.0.5 so make sure the 2.0.6 version is used.

@nvisionative or @valadas do you guys see anything else?

mean2me:
Thanks a lot for detailed answers @mtrutledge !
I'm already at work to implement changes you suggested.

david-poindexter:
This is a small thing, but I'd love to see us all agree on how we are going to handle **spellcheck/spell-checker**.

- I have been leaving it intact and just modifying `.eslintrc.js` to accommodate changes.  
- Other team members have been commenting it out or removing it altogether.
- With this PR, I am seeing a new approach emerge to use `.eslintskipwords.js`

Thoughts on how we should all be approaching this?

mtrutledge:
I have been setting spellcheck to false on the ESLint config for identifiers, but leaving comments, and strings in tact. The .eslintskipwords.js has been in the majority of projects I have touched. I have even added words to that file. If that file was missing and spellcheck was enabled it would have been throwing a ton of warnings. I would leave it for strings since there could be a string shown to the user. Enabling it for identifiers just causes a very large ignore file.

valadas:
It is my opinion that it should be removed, all user visible strings come from resource files that are not checked so we just take useless space and spend more time just adding skipwords with no benefit at all.

mean2me:
> I believe `CheckBox` should be `Checkbox`. Is it working the way it is now?

It worked, but I changed to Checkbox anyway in my last commit.

mean2me:
> The dnn-global-styles has moved to @dnnsoftware/dnn-react-common. Sorry if I did not mention that in my initial post.

No problem, just fixed in my last commit.

mtrutledge:
Thanks for the changes. Looks good!

mtrutledge (approved):
Looks good

ohine (approved):

mitchelsellers (approved):

PR #185, opened 2018-11-30
SEO: URL Regex: Settings are saved in HostSettings but pull…

mikebigun:
Once SEO Regex settings are saved/updated on a portal level and page reloaded, old settings are fetched from the cache. On update, portal cache should be cleared.


daguiler (approved):
LGTM

bdukes (approved):
LGTM

PR #198, opened 2018-12-04
Accepting spaces in search string.

mean2me:
This PR is related to https://github.com/dnnsoftware/Dnn.AdminExperience/pull/62, and fixes the following scenario:

Given below 6 users exist:

search one
search two
search three
user one
user two
user three

When search term is **search two**, the expected result is just one record (user **search two**), but we are getting 2 results (**search two** and **user two**).

Also, searching for **two search** currently yields 3 results (**search one**, **search two** and **search three**), which makes little sense. After this change, this search returns no results.


daguiler (approved):

daguiler:
@dnnsoftware/approvers please review

bdukes (approved):
LGTM

PR #238, opened 2018-12-16
Updates TaskScheduler module to use new react common components

### LINKED ISSUE #234: Task Scheduler module needs to be updated to use latest React Common components
## Description of bug
Task Scheduler module needs to be updated to use latest React Common components

## Current result
Module does not load

## Expected result
Module should load

## Additional context
I am working on that now

## Affected version

* [x] 9.3.0 (at the time of writing this)
* [ ] 9.2.1
* [ ] 9.2
* [ ] 9.1.1
* [ ] 9.1
* [ ] 9.0

### LINKED ISSUE #237: Collapsible component does not expand when editing Scheduled Tasks
## Description of bug
After updating the Scheduled Tasks module to use the latest react common components, when editing a task, the task editor renders in the html but it is hidden because its parent Collapsible component does not expand. The same Collapsible component expands fine when clicking the task history button.

## Steps to reproduce
1. Clone https://github.com/valadas/Dnn.AdminExperience/tree/issue-234
2. Try to edit a task
3. Notice the row does not expand
4. Inspect the html and notice the editor is rendered but in a div with a style of height:0px; overflor:hidden
5. Removing height:0px show everything fine

## Error log
No errors get log, something just does not remove the height:0px for some reason when expanding but only when clicking edit, it works fine when clicking on the history button

## Additional context
I tried hard to find the issue but I am stumped, will need some help.

## Affected version
https://github.com/valadas/Dnn.AdminExperience/tree/issue-234

## Affected browser

* [x] Chrome
* [?] Firefox
* [?] Safari
* [?] Internet Explorer
* [?] Edge

valadas:
## Summary
Updates TaskScheduler module to use new react common components

closes #234 

~~But: although everything seams to render fine, there is just a css issue with the height of the collapsible component. I will need help with #237~~

~~I think we can still merge this part and see after how to fix #237 with someone else helping out.~~


valadas:
So, I got it fixed the related issue on this commit. It ends up that Collapsible does not like the GridSystem and floats. So I replaced that by a simple css flexbox to do the columns and it works. 

This is ready for review now @mitchelsellers , @ohine , @mtrutledge

daguiler (approved):

mtrutledge (approved):
Good work @valadas

ohine (approved):

PR #247, opened 2018-12-19
PB > Security: bugfix and optimization

mean2me:
We are facing runtime errors on attempt to add a new ip filter due to wrong import name of Grid (that's now called GridSystem).
I also optimized build files and code.


Extensions/Settings/Dnn.PersonaBar.Security/Security.Web/src/components/ipFilters/index.jsx
```diff
@@ -42,17 +42,10 @@ class IpFiltersPanelBody extends Component {
         return <div className="header-row">{tableHeaders}</div>;
     }
 
-    onEnter(key) {
-        const { state } = this;
-        alert("You pressed enter! My value is: " + state[key]);
-    }
```
daguiler, line 48: 😄

daguiler (approved):

mtrutledge (approved):

PR #251, opened 2018-12-19
Greenkeeper/default/@dnnsoftware/dnn react common 2.0.8

### LINKED ISSUE #250: All package.json files need to be updated to use the latest version of the common components
## Description of bug
The common components are now v 2.0.8 and that needs to be updated in all the modules in the package.json, no breaking changes expected, just makes sure to use the latest ones to prevent build warnings and some ui styling issues.

This could be a single PR for all modules for simplicity.
Just need to run yarn upgrade @dnnsoftware/dnn-react-common@^2.0.8 on each project that uses React.

david-poindexter:
Fixes #250 

## Summary
Updates dnn-react-common to 2.0.8


valadas (approved):
Looks good to me

daguiler (changes requested):
I see myget source was changed in yarn.lock files

bdukes:
That's right @daguiler, we're pushing everything to the public npm registry now, and no longer depending on myget.

daguiler (approved):

Issue #245, opened 2018-12-19
Script bundles should be deleted and ignored


valadas:
If pull requests are not merged in order (often happens), there is always a merge conflict with the minified bundle script. This file gets generated by the build so it is not needed in the commits, it should be deleted from the repository and added to the gitignore file to simplify merging pull requests.

ohine:
Fixed in #255

valadas:
That got done in #255 Bye Bye Bundles,

PR #255, opened 2018-12-20
Adios bundles

ohine:
Removes the generated bundles from source control. This will help us know for a fact if a build is good, or bad right away. 

This fixes issue #245


valadas (approved):
Looks good to me

mtrutledge (approved):

PR #257, opened 2018-12-20
Fixed compilation errors in Extensions module

mikebigun:
Fix is similar to the one provided in https://github.com/dnnsoftware/Dnn.AdminExperience/pull/256
This is a small correction that fixes compilation warnings due to missing dependencies and deprecated settings in Extension module.


david-poindexter (approved):
Looks good to me.  Confirmed with @mtrutledge that it is okay to have **react** as a project dependency.  Webpack externals referenced in the webpack config ensures it is not bundled and that the references use the Common.

mitchelsellers (approved):

PR #261, opened 2018-12-23
Fixed UI issues in PB > Users > Roles

mean2me:
PersonaBar > Users > Roles was not displayed due to an error with debounce library. I couldn't make it work, so I choosed to remove it and adjust the code accordingly.

Build settings cleanup
CSS fixes
Code refactoring


valadas (approved):
Looks good to me and tested too. Can we get a second review please @dnnsoftware/ae-approvers

daguiler (approved):

PR #267, opened 2018-12-25
Fixed compilation errors in Themes, Servers and Vocabularies modules

mikebigun:
Few corrections are done to fix Themes, Servers and Vocabularies compilation warnings (due to missing dependencies and deprecated settings).
Fix is similar to the one provided in https://github.com/dnnsoftware/Dnn.AdminExperience/pull/256


valadas (approved):
Looks good to me

daguiler (approved):

PR #278, opened 2018-12-27
Removes spellchecking on Users module, fixes filter dropdown

valadas:
## Summary
Removes spellchecking on Users module and removes 2 eslint warnings

closes #277

PR #279, opened 2018-12-28
PB > Users > Search: restored Dropdown filter

mean2me:
Bugfix: Users filter dropdown disappeared due to variable's name typo.


valadas:
This fix is already included in #278

PR #80, opened 2018-09-18
Put check for undefined as boolean value was ignore either case

Mohtshum:
Fixes #79 

## Summary
Put check for undefined as boolean value was ignore in either case

PR #286, opened 2019-01-01
Get-portal &  get-site command blank screen

### LINKED ISSUE #285: DNN Products > Persona Bar > Settings > Prompt > get-portal displays blank screen & console errors
## Description of bug
get-portal and get-site are giving blank screen and errors in console of browser

## Steps to reproduce
1. Log in As Super User in Evoq Engage 
2. Go to Persona Bar > Settings > Prompt 
3. Type command get-portal/get-site 

## Current result
Application executes the command and blank screen is displayed. Console shown below error:
export-bundle.js?cdv=130:90 TypeError: Cannot read property 'toString' of null

## Expected result
Portal/Site information should be displayed. 

## Error log
export-bundle.js?cdv=130:90 TypeError: Cannot read property 'toString' of null

## Affected version
* [x] 9.3.0

Mhtshum:
Fixes #285 

## Summary
Put null check

### DIFF
--- a/Extensions/Settings/Dnn.PersonaBar.Prompt/Prompt.Web/src/utils/helpers.js
+++ b/Extensions/Settings/Dnn.PersonaBar.Prompt/Prompt.Web/src/utils/helpers.js
@@ -45,7 +45,7 @@ export function renderObject(data, fieldOrder) {
     const columns = !fieldOrder || fieldOrder.length === 0 ? getColumnsFromRow(data) : fieldOrder;
     const rows = columns.map((fldName, index) => {
         const lbl = formatLabel(fldName);
-        const fldVal = data[fldName] !== undefined && data[fldName] !== null ? data[fldName].toString() : "";
+        const fldVal = data[fldName] ? data[fldName].toString() : "";
         const cmd = data["__" + fldName] ? data["__" + fldName] : null;
 
         if (cmd) {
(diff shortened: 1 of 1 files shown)


daguiler (approved):

mitchelsellers:
@valadas Thoughts regarding this addition?  I believe your root cause analysis is correct in that this is due to the fact that we had that data issue.

valadas:
I also believe the root cause is the other issue, but I see no harm rechecking it in the module to prevent it from cashing in other situations.

In javascript, undefined is a variable that was declared but never assigned a value, null is when the value null was actually assigned to the variable. But both undefined and null coalesce to false, so:
```javascript
const fldVal = data[fldName] ? data[fldName].toString() : "";
```
would be a shorter way to do it.
I approve this change but will not merge in case @Mhtshum wants to make it shorter. But I will assign the 1.7.0 milestone to make sure we merge this in case @Mhtshum does not update to the shorter version. No big issue in either case.

valadas (approved):

daguiler:
Well, looking back in the history of the file, it was short in the beginning, and then @Mhtshum added the explicit undefined and the strict equality operator, and now the null. Not sure what was the motivation but it seems we are repeating ourselves. I also see no harm either way.

Mhtshum:
Thanks a lot @daguiler!, Actually in #80 the fix to display "false" values which was getting empty string because of short check. Hence mine initial check is correct for undefined & null.

daguiler:
@Mhtshum Can you please revert to short version as suggested by @valadas?

Mhtshum:
@daguiler and @valadas, As I mentioned short version is giving empty for attribute which have "false" hence undefined and null check is valid.
The issue reported for page which has isDeleted: false, in that case on prompt output we are getting empty result.

daguiler:
Ok, I see what you mean now @Mohtshum. Approving.

daguiler:
May I suggest adding a comment in code so that we don't stumble upon this again?

valadas:
Good catch @Mhtshum

PR #292, opened 2019-01-03
Updated UsersCommon Bundle

### LINKED ISSUE #291: Evoq Products like Engage is having old UsersCommon Bundle, which is causing issue for in PB > Manage > Users ( Clicking role icon on user blank screen)
## Description of bug
Evoq Products like Engage is having old UsersCommon Bundle, which is causing issue for in PB > Manage > Users ( Clicking role icon on user blank screen)

## Steps to reproduce
1. Login as Host
2. Go to PB > Manage > Users 
3. Click on Role icon for any registered user

## Current result
Blank screen

## Expected result
It should show dropdown to add role for the user

## Affected version
* [x] 9.3.0

Mhtshum:
- Need to verify how exportable Bundle under Users.Web will be updated for Evoq Products

Fixes #291


Mhtshum:
Though not sure how to this will get update in Evoq products builds.

valadas:
@ohine should we delete the bundle and get it to build by CI too ?

ohine:
Yes, the bundle should be deleted. Not sure how we missed this one.

daguiler (approved):
Approving meanwhile.

ohine (changes requested):
This file needs to be deleted. We are no longer storing any bundles and relying fully on the CI system to generate these.

daguiler:
My understanding is that this bundle requires an additional `yarn build` that is not yet implemented. Can you clarify @Mhtshum ?

ohine:
That's why we need all generated content removed so we can identify these issues quickly and fix them. We will not be accepting any pull requests which are updating items which should be handled by the CI system.

Mhtshum:
@daguiler you right, and the purpose of this PR was to highlight issue towards building of Sub React applications, indeed CI should take care of it by adding step mentioned by @daguiler

valadas:
I agree with @ohine the correct fix here is a pull request that would adjust the build scripts to account for those exportables, they should build when doing a release build, the CI will then just fire that up like normal.

valadas:
@Mhtshum , @ohine please take a look at #297 and let me know what you think.

daguiler:
Let's close this in favor of #297

PR #325, opened 2019-01-17
Extensions module: js errors on extension installation

mikebigun:
I found set of warnings on `yarn build` and multiple errors during extension installation in debugger (related to required props).

This PR is an addition to the main fix provided here  https://github.com/dnnsoftware/Dnn.React.Common/pull/203


mtrutledge (changes requested):
Instead of request default values after the props like `deleteExtensionsFiles || false` it would be better to set the `defaultProps` property for the component. Something like this:

```
class Control extends React.Component {

  // ...

  render() {
    // Do render
  }

}
Control.defaultProps = {value: ''};
```

mtrutledge (approved):
Awesome! Looks good.

valadas (approved):
Look good

PR #342, opened 2019-01-21
disableHostCheck during development

### LINKED ISSUE #324: Disabling host/origin check for development
## Description of bug
Due to recent security fixes with webpack-dev-server, we get a lots of host/origin warnings/errors when developing using yarn webpack. See this issue https://github.com/webpack/webpack-dev-server/issues/1604

## Additional context
This is not rush and can go for 9.4.0, but we would need to add ```disableHostCheck: true``` and make sure that this only applied to development build and not production.
Ideally it would be nice to have production and development webpack settings in separate files as it is shown in most current webpack examples on their site, one common file that load a dev or prod version in addition to the common settings, this would be easier to manage and more readable.

## Affected version

* [x] 9.3.0 RC0
* [ ] 9.2.1
* [ ] 9.2
* [ ] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser
I see this in Chrome, but I guess all modern browsers are affected.

valadas:
After the last update of webpack, there are a lot of errors in the browser console due to this issue https://github.com/webpack/webpack-dev-server/issues/1604

The current recommended way is to add a disableHostCheck in the devServer settings just for development. This PR implements that in all react projects but only if not build in production mode.

Closes #324 or maybe it does not and we close it manually again :)

### DIFF
--- a/Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/webpack.config.js
+++ b/Extensions/Content/Dnn.PersonaBar.Pages/Pages.Web/webpack.config.js
@@ -11,6 +11,9 @@ module.exports = {
         filename: "pages-bundle.js",
         publicPath: isProduction ? "" : "http://localhost:8080/dist/"
     },
+    devServer: {
+        disableHostCheck: !isProduction
+    },
     module: {
         rules: [
             {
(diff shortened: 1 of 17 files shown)


mitchelsellers (approved):

david-poindexter (approved):
Nice - thank you @valadas!

PR #337, opened 2019-01-19
Site Settings > Site Behavior > Site Aliases: Edit Alias behaves as Add Alias

anishmishra978:
Fixes #336

PR #343, opened 2019-01-21
Fixes an issue where it was impossible to edit portal aliases

### LINKED ISSUE #336: Site Settings > Site Behavior > Site Aliases: Edit Alias behaves as Add Alias
## Description of bug
ISSUE 1: the table with existing aliases has no headers
ISSUE 2:  Site Alias textbox is empty. Because in fact, Edit Alias is working as Add Alias

## Steps to reproduce
##ISSUE 1: the table with existing aliases has no headers
1) Navigate to PB > Site Settings > Site Behavior > Site Aliases
2) Press F5 to refresh

##ISSUE 2:  Site Alias textbox is empty. Because in fact, Edit Alias is working as Add Alias
1) Navigate to PB > Site Settings > Site Behavior > Site Aliases
2) Press F5 to refresh
3) Click Add Alias again
4) Add a new alias, for example platform-latest.dnndev.me. Leave all other fields with default values.
5) Click Save
6) Click the pencil icon to edit the newly created alias

## Current result
1. The table with existing aliases has no headers.
image

2. Site Alias textbox is empty. Because in fact, Edit Alias is working as Add Alias
image

## Expected result
1. header should be present for site Aliases table.
image

2. Site alias should be editable
image
## Affected version
* [ ] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser
* [x] Chrome
* [x] Firefox
* [x] Safari
* [ ] Internet Explorer
* [ ] Edge

valadas:
## Summary
This is a replacement for PR #337  and closes #336 

It fixes some build warnings about invalid prop types, fixes the editing of the site aliases (the deprecated componentWillReceiveProps event was deleted and not reimplemented in any way) It also fixes a UI issue where in 9.2.2 you would see a blank row -     -     -      -  with dashes after adding or cancelling adding an alias.

### DIFF
--- a/Extensions/Settings/Dnn.PersonaBar.SiteSettings/SiteSettings.Web/src/components/siteAliasSettings/siteAliases/index.jsx
+++ b/Extensions/Settings/Dnn.PersonaBar.SiteSettings/SiteSettings.Web/src/components/siteAliasSettings/siteAliases/index.jsx
@@ -183,7 +183,7 @@ class SiteAliasesPanel extends Component {
                     </div>
                     <div className="alias-items-grid">
                         {this.renderHeader()}
-                        <Collapsible isOpened={opened} style={{width: "100%", overflow: "visible" }}>
+                        <Collapsible isOpened={opened} style={{width: "100%", overflow: opened ? "visible" : "hidden"}}>
                             <SiteAliasRow
                                 alias={"-"}
                                 browser={"-"}
--- a/Extensions/Settings/Dnn.PersonaBar.SiteSettings/SiteSettings.Web/src/components/siteAliasSettings/siteAliases/index.jsx
+++ b/Extensions/Settings/Dnn.PersonaBar.SiteSettings/SiteSettings.Web/src/components/siteAliasSettings/siteAliases/index.jsx
@@ -145,7 +145,7 @@ class SiteAliasesPanel extends Component {
                         browser={item.BrowserType}
                         skin={item.Skin}
                         language={item.CultureCode}
-                        isPrimary={item.IsPrimary}
+                        isPrimary={!!item.IsPrimary}
                         deletable={item.Deletable}
                         editable={item.Editable}
                         showLanguageColumn={this.props.siteAliases.Languages && this.props.siteAliases.Languages.length > 1}
--- a/Extensions/Settings/Dnn.PersonaBar.SiteSettings/SiteSettings.Web/src/components/siteAliasSettings/siteAliases/index.jsx
+++ b/Extensions/Settings/Dnn.PersonaBar.SiteSettings/SiteSettings.Web/src/components/siteAliasSettings/siteAliases/index.jsx
@@ -189,7 +189,7 @@ class SiteAliasesPanel extends Component {
                                 browser={"-"}
                                 skin={"-"}
                                 language={"-"}
-                                isPrimary={"-"}
+                                isPrimary={false}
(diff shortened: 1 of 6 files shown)


Extensions/Settings/Dnn.PersonaBar.SiteSettings/SiteSettings.Web/src/components/siteAliasSettings/siteAliases/index.jsx
```diff
@@ -141,7 +145,7 @@ class SiteAliasesPanel extends Component {
                         browser={item.BrowserType}
                         skin={item.Skin}
                         language={item.CultureCode}
-                        isPrimary={item.IsPrimary}
+                        isPrimary={!!item.IsPrimary}
```
david-poindexter, line 148: This does not seem correct.  Was it intentional to use a double `!` operator?  Is there a situation where `item.IsPrimary` is 0, null, undefined, etc.?

daguiler: oh it's an old trick to convert to boolean

david-poindexter: Seems like a bit of a hack, but if that is considered best practice here, then fine.  Otherwise, I'd suggest tracing this down so that `item.IsPrimary` is never falsey.

daguiler (approved):
verified to work on latest build.

david-poindexter (approved):
Approving so as not to block for the `!!` "trick" to coerce `item.IsPrimary` to a boolean if falsey.  I do suggest this is eventually looked into though.

daguiler:
One comment though, probably not related to this PR:
1) fresh install, only 1 site alias (primary)
2) add alias "platform.com", click `Set Primary` button

Primary alias is now "platform.com", and now there is no way to set the initial alias as primary any more. Ever.

Extensions/Settings/Dnn.PersonaBar.SiteSettings/SiteSettings.Web/src/components/siteAliasSettings/siteAliases/index.jsx
daguiler: In minified javascript you'll notice many of these tricks. For instance, `true` is `!0`, and `false` is `!1`. But yeah, best to rely on type checking.

valadas: > Approving so as not to block for the `!!` "trick" to coerce `item.IsPrimary` to a boolean if falsey. I do suggest this is eventually looked into though.

It also prevents changes from elsewhere to affect the code here ! inverterts the coalesced bool and ! reverts it back so we are sure that at least inside the component we get a real bool that way. 

The problem with languages that do not have strong typing is that even with propTypes and stuff, it just gives a warning, and sometimes only in production, so this protects a bit against bad usage in the api or the component that uses this one, etc.

Would typescript prevent passing a wrong type to a component or would it also just throw a warning that is easy to ignore ?

PR #364, opened 2019-01-25
Fixes user email notification when registration type is none

### LINKED ISSUE #363: Cannot notify user about registration when registration type is none
## Description of bug
When the registration type is none, creating a user and ticking the checkbox to notify him does not send an email to the user.

## Steps to reproduce
1. In Security -> Member Accounts, set the registration type to none
2. In Users, create a new user and tick the "Send an email to the new user" checkbox

## Current result
The admin receives an email but not the user

## Expected result
The user should receive the email

## Additional context
This was previously discussed here https://github.com/dnnsoftware/Dnn.Platform/issues/2424 and an inital PR #2492 was merged to bring the logic in to the core but the corresponding UI part needs to be updated to use this new method signature. The signature just accepts one additional bool parameter to notify the user or not and already checks the registration mode of the site to decide which email to send.

## Affected version

* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [?] 9.1.1
* [?] 9.1
* [?] 9.0

valadas:
## Summary
Fixes user email notification when registration type is none.

Closes #363 

Please read issue and related issues to understand this change. Feel free to reply if not clear.

### DIFF
--- a/Extensions/Manage/Dnn.PersonaBar.Users/Components/RegisterController.cs
+++ b/Extensions/Manage/Dnn.PersonaBar.Users/Components/RegisterController.cs
@@ -240,7 +240,7 @@ namespace Dnn.PersonaBar.Users.Components
             newUser.Membership.PasswordQuestion = registerationDetails.Question;
             newUser.Membership.PasswordAnswer = registerationDetails.Answer;
             //final creation of user
-            var createStatus = UserController.CreateUser(ref newUser);
+            var createStatus = UserController.CreateUser(ref newUser, registerationDetails.Notify);
 
             //clear cache
             if (createStatus == UserCreateStatus.Success)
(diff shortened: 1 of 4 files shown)


mitchelsellers (approved):

daguiler (approved):
Hi @valadas, I tested all user registration modes and everything looks good. I got the correct emails every time. So approving.

However, I spotted something that's not strictly related to this PR, but has to do with the end-to-end use case:

When a new user is added **manually** by the host (not via self registration), AND the user registration mode is Private, the user gets an email with following text:
> Your account details will be reviewed by the website Administrator and you will receive a notification upon account activation.

This is not accurate because users are automatically approved when manually added by the host, so no further notification is sent. I think we should send the Public registration message instead:
> We are pleased to advise that you have been added as a Registered User to My Website.

If the site's registration mode is Public, then the above issue does not occur. The second message is sent in the email. So when users are added manually by the host, we are sending one message or the other depending on the user registration mode, which is wrong. We should always send the second one.

daguiler:
Created #367 for this.

valadas:
> Created #367 for this.

Makes sense

PR #374, opened 2019-01-31
Fixes error when saving SSL settings

### LINKED ISSUE #373: Error saving SSL settings
## Description of bug
Under certain circumstances when saving SSL settings it's possible to get duplicate primary key error on PortalAlias table.

## Steps to reproduce
1. Open PB>Settings>Security>More>SSL Settings
2. Set SSL URL to `https://some.domain.com`
3. Click Save
4. Change SSL URL again to `https://some.domain.com/` (notice the trailing slash)
5. Click Save

## Current result
HTTP 500 response on /API/PersonaBar/Security/UpdateSslSettings
```
{"Message":"An error has occurred."}
```

## Expected result
Save should succeed normally.

## Affected version

* [x] 9.3.0

daguiler:
Fixes #373

## Summary
The error described in #373 occurs because of an attempt to insert a duplicate value in the PortalAlias table in the context of saving SSL settings. This occurs because DNN fails to recognize that the alias already exists when aliases only differ in trailing slashes.

In addition to fixing above issue, this PR includes a small UI modification to refresh the SSL Settings page after save. This is necessary because the save logic might alter the values specified by the user. For example, the values entered for SSL URL and  Standard URL are trimmed, converted to lowercase, the protocol part is removed, and now the trailing slash is also removed.


daguiler:
dnn-28315-coverage

valadas (approved):
Looks good and works

ohine (approved):

Issue #381, opened 2019-02-04
DNN Platform RC1 > Import/Export > User is unable to select Export package on Import Data screen


DinKumar:
## Precondition
1. Latest DNN Platform is installed 
2. Host user credentials are available 
3. Child site is added 
4. Full export package is created for Default website

## Description of bug
User can't select the Export Package when user clicks on the exported package on Import data screen.

## Steps to reproduce
1. Login to DNN Platform with host user credentials 
2. Go to Persona Bar > Settings > Import / Export 
3. Select Child Site from Site drop down list 
4. Click on Import Data button 
5. On Import Data Screen, select the package. (Exported package displayed here) 

## Current result
Package is not getting selected when user clicks on the exported package.

## Expected result
Export package should get select and user should be able to continue with importing the package into child site.

## Screenshots
![import-export_selectpackage](https://user-images.githubusercontent.com/40901826/52235419-056d0580-28ea-11e9-9561-a9636535fbf2.png)

## Error log
Paste the error log that is related to this issue.

## Additional context
Add any other context about the bug that may be helpful for resolution.

## Affected version

* [x] Platform RC1
* [ ] 9.2.2
* [ ] 9.2.1
* [ ] 9.2
* [ ] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser

* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

valadas:
Closing, PR merged

PR #388, opened 2019-02-07
DNN Platform > Import/Export > User is unable to select Exp…

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.AdminExperience/issues/381

I found lot of react migration problems that causes console errors. All of them are fixed here.


ohine (approved):

valadas (approved):
I could not reproduce the original issue before this PR but those changes look good anyway.

daguiler:
Same here. Just in case I tested this and it works fine.

PR #394, opened 2019-02-08
Placeholder to avoid PersonaBar slide effect

### LINKED ISSUE #393: Persona Bar slide effect on every refresh
## Description of bug
On every roundtrip to the server, the PersonaBar menu is loaded again. It usually takes a few moments to load and it remains hidden until it does. Then it shows with this sliding effect that moves the entire page to the right. This is very annoying and may even make the user click on the wrong UI components.

## Steps to reproduce
1. Login to DNN as a host
2. Observe the PersonaBar slide effect when it loads.

## Current result
The whole page is moved to the right.

## Expected result
It would be a much better UX if the page didn't move all the time because of the PB.

## Affected version

* [x] 9.3.0
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [ ] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser

* [X] Chrome
* [X] Firefox
* [X] Safari
* [X] Internet Explorer
* [X] Edge

daguiler:
Fixes #393

## Summary
This pull request creates a temporary placeholder for the PersonaBar that is immediately rendered on the page, preventing the sliding effect that occurs every time PB loads.

~This PR depends on https://github.com/dnnsoftware/Dnn.Platform/pull/2583, which needs to be merged first.~


valadas:
I had something else in mind I wanted to submit, so please wait to compare both solutions to see what people like most. I am trying to submit that today.

valadas:
After some digging around, I did not like my alternate solution, so I am all for this as a temporary fix for the sliding animation annoyance until we come up with something better. @daguiler if you can just move the CSS from the other PR in here, I would approve that.

daguiler:
Moved CSS required by PB placeholder to `personaBarContainer.css`.

valadas (dismissed):
Looks good to me, tested in Chrome, Firefox and Edge.

I think this is a good workaround until we have more time to do something better about the PersonaBar loading, any chances we can get this into the current RC? I do not see any risk in the changes and it would make many people happy to stop the moving around of page controls.

ohine (dismissed):
Looks good as an interim solution to loading delay issues with the sliding animation. Thanks @daguiler!

ohine:
The risk is very minimal, but we are way past the code freeze date so it's a tricky line. I also completely agree it'll help, but on the other hand I also see the value in sticking to our published code freeze policy as well. 

I approved the PR but didn't merge as we need the rest of the group to make a decision on which version it'll go into.

bdukes:
As much as I'd love to have it in, I think it's too late in the process, and needs to be a part of 9.3.1

mitchelsellers:
I concur that I think this needs to become a 9.3.1 item, unless we have a true need to extend out a period of extended RC validation

daguiler:
Hi all, sorry to make last minute changes. I just converted the placeholder DIV into a server side Panel, for better extensibility. It doesn't change anything but it enables the panel to be found and customized server side.

ohine:
@daguiler Is there anyway we can not use a serverside control? It adds to the payload quite a bit.

daguiler:
Ok, yes, there is another way. Will submit shortly. Actually I'll revert last commit and find a different way for customizing.

ohine:
> I concur that I think this needs to become a 9.3.1 item, unless we have a true need to extend out a period of extended RC validation

@mitchelsellers, we have at least one more RC, it was due last week but we still have bugs the group wants fixed before generating another RC.

ohine:
@daguiler I'm curious, Could we just switch things around slightly to use `personabar-placeholder` instead of `personalBarContainer`? And then just remove the new panel and adjust the css that's on the main placeholder?

daguiler:
I'm sure it's possible, but it's hard to grasp the exact changes you are describing. Maybe best way is to create a parallel PR with your proposed changes?

mtrutledge:
Are there any other changes to be made for this? It seems the last comment had a suggestion of a change.

valadas:
@dnnsoftware/ae-approvers Should I click the Update Branch button and see if this still builds fine then merge it ?

daguiler:
Will see if I can save this PR. Re-targetting brings unwanted commits

ohine:
If it’s too much trouble, we can always just wait for 940 and merge this into that branch.

daguiler:
The PR cannot be saved if it was updated with latest changes before changing target branch, which makes sense. I tried to force-push my branch to see if the "update" could be reverted, but now Github refuses to work with this PR anymore. I'll create a new one.

PR #401, opened 2019-02-13
API Call 500 Error For Username Already Exists

### LINKED ISSUE #383: DNN Platform RC1 > Persona Bar > Manage > Users> Console Errors on Add User with existing username
## Precondition
1. Install Platform RC1
2. Host user credentials are available 
3. Authorized user with email id as username OR normal username is created
## Description of bug
Application throws 500 (Internal Server Error) while adding user with existing username.

## Steps to reproduce
1. Login to DNN Platform as host user /Administrator user 
2. Go to Persona Bar > Manage > Users 
3. Click on Add User button 
4. Fill valid inputs on Add User screen except User Name-> Add existing username 
5. Click on Save button

## Current result
Application displays error message which is expected but console error is also displayed: 
POST http://dw1.eu.aureacentral.com:21030/API/PersonaBar/Users/CreateUser 500 (Internal Server Error) 

## Expected result
Application displays error message for username is already in use and there should not be any console error.

## Screenshots
consoleerrorforexistinguser

## Error log
jquery.js?cdv=52:11 POST http://168.61.157.87/API/PersonaBar/Users/CreateUser 500 (Internal Server Error)

## Additional context
Add any other context about the bug that may be helpful for resolution.

## Affected version

* [x] Platform RC1
* [ ] 9.2.2
* [ ] 9.2.1
* [ ] 9.2
* [ ] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser

* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

Mhtshum:
Fixes #383 

## Summary
- Throwing SecurityException in RegistrationController
- Converting response to BadRequest for SecurityException

### DIFF
--- a/Extensions/Manage/Dnn.PersonaBar.Users/Components/RegisterController.cs
+++ b/Extensions/Manage/Dnn.PersonaBar.Users/Components/RegisterController.cs
@@ -30,6 +30,7 @@ using Dnn.PersonaBar.Library;
 using Dnn.PersonaBar.Library.Common;
 using Dnn.PersonaBar.Users.Components.Contracts;
 using Dnn.PersonaBar.Users.Components.Dto;
+using Dnn.PersonaBar.Users.Components.Exceptions;
 using DotNetNuke.Common;
 using DotNetNuke.Common.Utilities;
 using DotNetNuke.Entities.Portals;
@@ -104,7 +105,7 @@ namespace Dnn.PersonaBar.Users.Components
             //ensure this user doesn't exist
             if (!string.IsNullOrEmpty(username) && UserController.GetUserByName(portalSettings.PortalId, username) != null)
             {
-                throw new SecurityException(Localization.GetString("RegistrationUsernameAlreadyPresent",
+                throw new BadRequestException(Localization.GetString("RegistrationUsernameAlreadyPresent",
                     Library.Constants.SharedResources));
             }
 
--- a/Extensions/Manage/Dnn.PersonaBar.Users/Components/RegisterController.cs
+++ b/Extensions/Manage/Dnn.PersonaBar.Users/Components/RegisterController.cs
@@ -24,6 +24,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Security;
 using System.Text.RegularExpressions;
 using Dnn.PersonaBar.Library;
 using Dnn.PersonaBar.Library.Common;
@@ -103,7 +104,7 @@ namespace Dnn.PersonaBar.Users.Components
             //ensure this user doesn't exist
             if (!string.IsNullOrEmpty(username) && UserController.GetUserByName(portalSettings.PortalId, username) != null)
             {
-                throw new Exception(Localization.GetString("RegistrationUsernameAlreadyPresent",
+                throw new SecurityException(Localization.GetString("RegistrationUsernameAlreadyPresent",
                     Library.Constants.SharedResources));
             }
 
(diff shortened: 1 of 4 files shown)


Extensions/Manage/Dnn.PersonaBar.Users/Components/RegisterController.cs
```diff
@@ -103,7 +104,7 @@ public UserBasicDto Register(RegisterationDetails registerationDetails)
             //ensure this user doesn't exist
             if (!string.IsNullOrEmpty(username) && UserController.GetUserByName(portalSettings.PortalId, username) != null)
             {
-                throw new Exception(Localization.GetString("RegistrationUsernameAlreadyPresent",
+                throw new SecurityException(Localization.GetString("RegistrationUsernameAlreadyPresent",
```
daguiler: Let's not use `SecurityException`, let's create a new exception type, for instance `BadRequestException` and throw that one instead.

daguiler (changes requested):
We cannot just fix this particular validation and leave all other validations as is. There are some other validations that will still return a 500 error. Rule should be:
* if the request cannot be processed because user supplied wrong arguments (this includes user already exists), then throw a BadRequest exception and then return a 400 HTTP response.
* in case of any unhandled exceptions, return a 500 response

Also please make sure that Prompt commands still work

Extensions/Manage/Dnn.PersonaBar.Users/Components/RegisterController.cs
Mhtshum: > We cannot just fix this particular validation and leave all other validations as is. There are some other validations that will still return a 500 error. Rule should be:
> 
> if the request cannot be processed because user supplied wrong arguments (this includes user already exists), then throw a BadRequest exception and then return a 400 HTTP response.
> in case of any unhandled exceptions, return a 500 response
> 
> Also please make sure that Prompt commands still work

Currently Prompt Command for existing username is not throwing any http error in browser console. i.e. 400

Mhtshum: - New exception type added 
- All other exceptions are already return with 500 response
- Prompt Command for 'new-user' is not returning any http error response for existing user

Extensions/Manage/Dnn.PersonaBar.Users/Services/UsersController.cs
```diff
@@ -25,6 +25,7 @@
 using System.Net;
 using System.Net.Http;
 using System.Web.Http;
+using System.Security;
```
daguiler: This can probably be removed now

Extensions/Manage/Dnn.PersonaBar.Users/Components/RegisterController.cs
daguiler: I see 9 more cases to fix where we are throwing `Exception` and we should be throwing `BadRequestException`

daguiler: @Mhtshum ^^

ohine:
I believe this was changed a few years ago to harden the system due to a few of the security issues and information leakage scenarios. I’d be happy to chat more about it in a private channel.

valadas:
Let me know after discussed if this needs change or needs being closed.

valadas:
@dnnsoftware/tag what is the plan on this, it is by design and this issue is to be closed ?

ohine:
@valadas We need this added to our next reviewers meeting agenda, so we can discuss and make a final decision on this PR.

mitchelsellers:
Closing this issue, as the generic exception that is thrown in this situation has been done on purpose.

Issue #403, opened 2019-02-13
Scheduler: Blank page is displayed when we click on "X" icon of Scheduler date and time


mikebigun:
## Description of bug

Resetting scheduler date/time value causes js errors with a blank screen.

## Steps to reproduce

Login as a SuperUser 
1. Navigate to Persona Bar -> Mouse hover on Settings -> Click on Scheduler 
2. Click on Scheduler Tab 
3. Click on `Edit` (pencil icon) of any task 
4. Select date and time from date picker for Schedule Start Date/Time field 
5. Click on `Apply` button 
6. Click on `X` icon of Schedule Start Date/Time

## Current result
Blank screen along with js error in console. 

## Expected result
Selected date and time from date picker should be removed from the screen on clicking on `X` icon and default value i.e. `MM/ DD / YYYY hh : mm AM` should be displayed

## Affected version

* [x] 9.3.0

valadas:
closing, PR fixed it

PR #404, opened 2019-02-13
Scheduler - resetting date/time value

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.AdminExperience/issues/403

See error below. To fix it, we should handle both `undefined` and `null` values.

image


daguiler (approved):
Works fine on latest code.

valadas (approved):
Tested working, changes look good

PR #413, opened 2019-02-26
Relative path of personaBarDialog.css

mikebigun:
This is a small correction of the relative `personaBarDialog.css` path. Relative path is configured globally and supplied with a `cssPath` variable (see [source code](https://github.com/dnnsoftware/Dnn.AdminExperience/blob/b9bda245e8381811733b897cfae6153f26a55203/EditBar/Dnn.EditBar.UI/editBar/scripts/main.js#L11)).

After the change, it works as expected: 

image


mitchelsellers (approved):
LGTM and important to support all portal types

daguiler (approved):

Issue #414, opened 2019-02-27
Page localization settings tooltips


mikebigun:
## Description of bug
Users are experiencing some problems when configuring module localization settings. To prevent confusion, need to add more explanation to the page localization tab. Good to have tooltips added and provide details what particular setting does.

## Steps to reproduce
1. Go to `Settings -> Site Settings`
2. Go to `Languages`
3. Add one or more languages, i.e. fr-FR, it-IT
4. Enable Content Localization
5. Translate all default pages 
6. Go to `Home` Page settings
7. Go to `Localization` Tab

## Current result
Currently, page localization tab contains set of modules configurations (checkboxes and links) which are not well-descriptive. It is not always clear what all of these settings means. Due to lack of information it causes confusion and after the changes users are experiencing unexpected behavior, i.e. when localizing or de-localizing modules.

## Expected result
Would be good to have tooltips on hover describing what particular checkbox/link does. That will help users to make proper configurations. See screenshot below with an expected result. 

## Screenshots
![image](https://user-images.githubusercontent.com/22524011/53508855-41e3e980-3ac3-11e9-8fa7-2583a50acdcd.png)

## Affected version

* [x] 9.3.0
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

valadas:
Closed by #723

PR #415, opened 2019-02-27
Page localization settings tooltips

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.AdminExperience/issues/414

See result below. This will help customers a lot when configuring page/module settings.

image


mitchelsellers:
@dnnsoftware/tag Thoughts on including this in 9.3.0?  Not a high-risk item, but don't want to unnecessarily add changes at this stage of the game

iJungleboy:
@dnnsoftware/tag IMHO it looks trivial, but I would like to kind of "standardize" that we don't add out-of-scope micro-improvements when a release is already rolling. These things tend to backfire - so rather for 9.3.1 or 9.4.0

sleupold:
I agree with @iJungleboy, especially as this is an enhancement, not a fix and the code modification is not trivial (would require a new RC)

bdukes:
Agreed, this should be held for 9.3.1

valadas:
It looks trivial and I love it, but just because we are already on RC2 I would hold it for 9.3.1 to be extra safe.

mikebigun:
Dear community, I'm closing this PR in favor of https://github.com/dnnsoftware/Dnn.AdminExperience/pull/435 which is targeted to `development`.

Issue #420, opened 2019-02-28
DNN Platform RC2 > Persona Bar > Manage > Users>Add User throwing error messages


DinKumar:
## Precondition
1. Install DNN Platform RC2
2. Host user credentials are available

## Description of bug
Below issues are observed for Add User feature:
1. When host user clicks on Add User  button then Authorized toggle button is displayed as Off.
2. Application is showing "Email is required" error message for valid email id 
3. Reenter same email id again and fill all required data then application displays error message "The username specified is invalid. Please specify a valid username."
4. Console error is displayed as "Failed to load resource: the server responded with a status of 500 (Internal Server Error) http://168.61.157.87/API/PersonaBar/Users/CreateUser:1"

## Steps to reproduce

1. Login to DNN Platform as host user /Administrator user 
2. Go to Persona Bar > Manage > Users 
3. Click on Add User button 
4. Fill valid inputs on Add User screen
5. Click on Save button

## Current result
Application displays error for valid email id, valid userid and console log shows error.

## Expected result
User with valid inputs should be added without any error messages. There should not be any console error for adding user function.

## Screenshots
![platform-addusererror-01](https://user-images.githubusercontent.com/40901826/53558310-c68b4200-3b6d-11e9-81aa-10def15361d5.png)
![platform-addusererror-02](https://user-images.githubusercontent.com/40901826/53558311-c68b4200-3b6d-11e9-8760-58e9d738f595.png)

## Error log
2019-02-28 09:53:23.729+00:00 [dnn-ci][D:2][T:25][ERROR] Dnn.PersonaBar.Users.Services.UsersController - System.ArgumentException: The username specified is invalid. Please specify a valid username.
 at Dnn.PersonaBar.Users.Components.RegisterController.Register(RegisterationDetails registerationDetails)
 at Dnn.PersonaBar.Users.Services.UsersController.CreateUser(CreateUserContract contract)

## Additional context
Add any other context about the bug that may be helpful for resolution.

## Affected version

* [x] 9.3.0 RC2
* [ ] 9.2.2
* [ ] 9.2.1
* [ ] 9.2
* [ ] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser

* [x] Chrome
* [ ] Firefox
* [ ] Safari
* [ ] Internet Explorer
* [ ] Edge

sleupold:
[...]

PR #421, opened 2019-02-28
DNN Platform RC2 > Users> Add User throwing error messages

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.AdminExperience/issues/420

Rootcause:
There is a regression found in https://github.com/dnnsoftware/Dnn.AdminExperience/pull/406
After that fix, on password change, `userDetails` object gets completely cleared, so only `password` property gets assigned to it. This leads to incorrect behavior when rendering and validating fields on a save.

To satisfy both browsers chrome and edge, I did a small correction to keep `UserDetails` refreshed only with a new password value. All other properties remains unchanged.

See short demo (tested on chrome and edge) : [demo](https://drive.google.com/file/d/1O629Q5p2MrlyoP966mDPlYCdvgG4evun/view?usp=sharing)


daguiler (approved):
Works fine in Chrome, could not test in Edge.

valadas (approved):

Issue #436, opened 2019-03-19
Persona Bar Pages goes blank when no EDIT permission


mikebigun:
## Description of bug
Persona Bar Pages goes blank when no EDIT permission.
This issue was observed in Evoq edition having extended list of permissions, but the root cause of the issue we found in a core PB part. 

## Steps to reproduce

1. As host, create a custom role named `Custom Role`
2. Create a new user named `Custom User` 
3. Add `Custom User` to `Custom Role` 
4. Create a page named "Custom Page", allowing "Custom Role" the `VIEW`, `ADD` and `ADD CONTENT` permissions 
5. Publish the page 
6. Log in as `Custom User` and navigate to `Custom Page`
7. Open PB > Content > Pages

## Current result
Js error and blank pages screen.

## Expected result
Pages PB panel should show up correctly.

## Error log
`Uncaught TypeError: Cannot read property 'includeInMenu' of null
    at eval (PersonaBarPageTreeviewInteractor.jsx?5cd8:61)
    at eval (App.jsx?c58c:884)
    at Array.forEach (<anonymous>)
    at left (App.jsx?c58c:883)
    at loop (App.jsx?c58c:889)
    at ProxyComponent._traverse (App.jsx?c58c:894)
    at ProxyComponent._traverse (VM42155 react-hot-loader.development.js:587)
    at ProxyComponent.UNSAFE_componentWillReceiveProps (PersonaBarPageTreeviewInteractor.jsx?5cd8:56)
    at ProxyComponent.UNSAFE_componentWillReceiveProps (VM42155 react-hot-loader.development.js:587)
    at callComponentWillReceiveProps (react-dom.development.js?61bb:12567)`

## Affected version

* [x] 9.3.0
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

valadas:
PR Merged

PR #437, opened 2019-03-19
Persona Bar Pages goes blank when no EDIT permission

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.AdminExperience/issues/436

There are multiple corrections:
1. I reverted some changes for `PersonaBarPageTreeviewInteractor` I did in 5fc4c549e2fc8582691b4c3001ee6028535103de previously, I found that was not a correct solution
2. Deprecated life-cycle methods are replaced with a new ones
3.  `includeInMenu` property is fixed to check whether it is undefined or not

Pages treeview works fine now in both Platform and Evoq.
Also, without `EDIT` rights user is not able to modify page settings but is possible to add child pages and add new modules to the page:

image


sleupold:
Permissions should not just check for "Edit" and "view", but should support granular permissions like in Evoq as well (in case a 3rd party permission provider is used).

daguiler:
So @mikebigun there is no actual change in permissions behavior in this PR correct? It's all about fixing front-end React bugs

mikebigun:
Right, only UI.

mikebigun:
Dear approvers, just want to follow up this one. Let me know if you have any concerns.

daguiler (approved):

valadas (approved):

PR #438, opened 2019-03-19
Add Dnn.React.Common into this repository

### LINKED ISSUE #375: Merging Dnn.React.Common into Dnn.AdminExperience
## Description of problem
Merging Dnn.React.Common into Dnn.Admin experience would help simplify developement.

## Description of solution
We need to move the files into this repository in it's own folder and do adjustments to CI maybe.

## Additional context
This is the result of the discussions in https://github.com/dnnsoftware/Dnn.Platform/issues/2436 and https://github.com/dnnsoftware/Dnn.Platform/issues/2556 and it appears we have a consensus on this part.

It may required adjustment to CI and https://github.com/nbrane/Dnn.DevelopmentConfig

bdukes:
This change merges the history of the two repositories together, and moves the contents of the Dnn.React.Common repository into a `Dnn.React.Common` folder.

@ohine not sure what changes need to happen in the Azure DevOps pipeline based on this change.
cc: @mtrutledge @valadas 

Fixes #375


ohine:
We should rework the AE solution to trigger the build commands for the react.common components. Then we would only need to handle the publishing of the package in Devops.

bdukes:
FYI, I've updated this branch so that it incorporates #439

valadas:
@ohine after this is merged, can we reference the common components directly instead of the nuget package and only release to nuget on actual dnn releases? Would that make sense?

valadas:
Also, are we targeting this for the 2.0.1 release or 2.1.0 or 3.0.0 ? Just to make sure we target the right branch and trying to avoid unexpected problems that would delay a quick 2.0.1 bug fix release?

bdukes:
@valadas I would think you would use `yarn link` to reference the local version of the package.  

I'm expecting that this change would target 2.1.0.

valadas:
The plan was to use yarn workspaces at some point (which uses symlinks under the hood) but on top of making debugging easier, would also deduplicate the packages used in each module. 

Back to the packages, I said nuget above, but meant npm packages. If we can reference them directly then we would not need to have them published to npm on each PR to use them in the modules, so we could only release to npm when we are done with a specific release right? (Which was one of the goals of merging these repositories).

bdukes:
Yes, workspaces should allow that

ohine:
> Also, are we targeting this for the 2.0.1 release or 2.1.0 or 3.0.0 ? Just to make sure we target the right branch and trying to avoid unexpected problems that would delay a quick 2.0.1 bug fix release?

If we can get this completed, I don't see an issue merging it into v2.0.1.

> The plan was to use yarn workspaces at some point (which uses symlinks under the hood) but on top of making debugging easier, would also deduplicate the packages used in each module.

Lets combine things first, and then we can add the yarn workspaces as a secondary pull request? Unless you think reworking yarn workspaces is a quick task. 

>  If we can reference them directly then we would not need to have them published to npm on each PR to use them in the modules, so we could only release to npm when we are done with a specific release right? (Which was one of the goals of merging these repositories).

Correct, that's my understanding as well.

mtrutledge:
I think we get it merged first and a build working and then switch over to yarn workspaces.

valadas:
`If we can get this completed, I don't see an issue merging it into v2.0.1.`

I am just scared to get into troubles that would delay a quick release cycle on 2.0.1...  I was under the impression we would try and only have bug fixes on 2.0.1, I see this more as an improvement and would personally keep that for 2.1.0 ? opinions ?

`I think we get it merged first and a build working and then switch over to yarn workspaces.`

Yes, we need this merged before implementing workspaces for sure...

mtrutledge (approved):
LG

valadas:
@ohine @mtrutledge @bdukes , In order to use yarn workspaces, we need the workspaces config file to be in a folder that is a parent to each project. Sooo, right now that config is in the WebApps folder and this would be a folder on the root of the project. What is the best plan here, do we merge this, close React.Common repo and make another PR to make it work ? Or is it better to redo this PR or do we need to get together and figure out a plan. I have no idea how Nuget publishing, is that some concern or it does not matter?

ohine:
I was planning on pushing commits into this PR to get things building. The first step would be to move the yarn workspaces configuration to the root of the repo and then adjust webpack for the updates paths. The piece I’m not quite sure on is how to get AE to consume the output from react.common since we won’t be using an npm package internally.

mtrutledge:
The yarn workspaces should be in the root of the entire project. To move away from the NPM packages we would need to not only use yarn workspaces but we also need to use a tool called Lerna. Lerna gets installed in the root package.json where you define the yarn workspaces. You then add a lerna.json file like the following

```
{
  "lerna": "2.11.0",
  "packages": [
    "packages/*"
  ],
  "version": "independent",
  "npmClient": "yarn",
  "useWorkspaces": true
}
```
This will essentially evaluate packages from npm and anything referenced in the packages definition through symlinks. Lerna does this automatically you just have to configure it. To get a better over view of yarn workspaces and lerna working together this is a good starting article:

https://medium.com/naresh-bhatia/sharing-ui-components-with-lerna-and-yarn-workspaces-be1ebca06efe

valadas:
Hmm, I thought yarn workspaces would have handled that without adding anything else...

mtrutledge:
While there is crossover between the functionality of yarn workspaces and Lerna, Lerna gives functionality and tools to maintain packages, versions, and other items for each of the items in the workspace. The two tools go hand in hand and Lerna will use yarn and yarn workspace commands when appropriate.

donker (approved):

ohine (approved):
Thanks for this contribution @bdukes! This is really going to speed up future development quite a bit! 

These changes look great, I've pushed a few commits which integrate everything into a single yarn workspace and ensure that Visual Studio triggers the correct commands when building the solution in Release mode.

PR #440, opened 2019-03-21
update version to v2.0.0

ohine:
Prior to v9.3.0, DevOps was setting both platform and AE versions at build time. Starting with v9.3.0 we switched to using GitVersion on platform. However, since we haven't yet configured GitVersion on AE the code in our DevOps task wasn't updating the version correctly.

This PR updates the various versions to their correct value. For v9.3.1/v9.4.0 we should have GitVersion wired up in AE so this won't need to be manually updated for future releases.


valadas (approved):
Looks good

bdukes (approved):

PR #649, opened 2019-04-05
replace jre based yui compressor with .net msbuild version

### LINKED ISSUE #377: Replace yuicompressor with modern tooling
## Description of problem
YUI Compresser needs java (JRE) to be installed on the build machine which is not always the case or even possible in some environments.

## Description of solution
It is currently used by these 3 build scripts: https://github.com/dnnsoftware/Dnn.AdminExperience/search?q=yuicompressor&unscoped_q=yuicompressor
We need to consider if is even required giving that most modules are already minified + if we ensure each module comes out minified, then we can skip that in the build process all together and shift that reponsibility into each module (most are react and use webpack, some are handlebars an not minified). If not possible we should consider replacing it with a solution that does not rely on JRE.

ohine:
## Summary
Closes #377


bdukes:
@ohine did you have a plan to address the build error:
>`Build\BuildScripts\Package.Targets`(68,5): Error : There is no data in the file [D:\a\1\s\Extensions\Manage\Dnn.PersonaBar.Roles\admin\personaBar\css\Roles.css]. Please check that this is the file you want to compress.
>`Build\BuildScripts\Package.Targets`(68,5): Error : Failed to read/parse data in file [D:\a\1\s\Extensions\Manage\Dnn.PersonaBar.Roles\admin\personaBar\css\Roles.css].
>`Build\BuildScripts\Package.Targets`(68,5): Error : Value cannot be null.
Parameter name: source

I took a look at it and wasn't able to find a way to suppress the error or add some placeholder content to the CSS file to quiet the warning.  Is Webpack already minifying, and the YUI step is unnecessary?  Or do we just need to not target this one file?

ohine:
@bdukes Yes, I was planning on circling back and getting this fixed, we definitely don't want to run YUI on the webapp folders, but I think we do want to run it on these files as yarn/webpack/etc isn't setup to process those folders. 

Do we even need the roles.css file if it's empty? Can we just delete it?

bdukes:
After removing files related to `Roles.css` and references, there's now this error:
>"C:\Code\Dnn.Platform\Dnn.AdminExperience\Dnn.AdminExperience.sln" (default target) (1) ->
"C:\Code\Dnn.Platform\Dnn.AdminExperience\Extensions\Manage\Dnn.PersonaBar.Roles\Dnn.PersonaBar.Roles.csproj" (default target) (17) ->
(Minification target) ->
  C:\Code\Dnn.Platform\Dnn.AdminExperience\Build\BuildScripts\Package.Targets(68,5): error : At least one file is required to be compressed / minified. [C:\Code\Dnn.Platform\Dnn.AdminExperience\Extensions\Manage\Dnn.PersonaBar.Roles\Dnn.PersonaBar.Roles.csproj]

So I think the MSBuild task still needs to be adjusted :-/

bdukes:
@ohine I pushed a couple of commits to address those errors 🤞

ohine:
@bdukes Thanks!! I was in the middle of wrapping up another issue before switching back to this! 🤞 for sure!

bdukes:
Looks like our checks are green, just need some approvals

bdukes (approved):

bdukes:
Upon review, I pushed one more change to comment out the JS minification pieces to clarify that they're unused.

mitchelsellers (approved):

mitchelsellers:
@ohine Want this merged now?

PR #788, opened 2019-04-13
Changes biography security check for a warning

### LINKED ISSUE #98: Update Security Analyzer to show a warning for RichText profile properties
## Description of problem
Currently the security analyzer flags the Biography field as an failure due to it's default configuration of being a rich text field. 

## Description of solution
We need to switch this to a warning, maybe update the steps to include a bit more details on where to go and change this to a basic text input.

valadas:
Closes #98 

## Summary
This changes the biography security check in the security analyser for a warning instead of a failure and updates the text to show where this setting can be changed. Note that with 9.3.2 this setting now defaults to plain text so in order to test, you need to change it to richtext first.

image


daguiler (approved):

david-poindexter (approved):
Looks good to me - thanks @valadas!

PR #791, opened 2019-04-13
Fixes not being able to select import package with description

### LINKED ISSUE #718: Unable to select site import package in 9.3.1
## Description of bug
Using DNN 9.3.1, cannot select import package that was exported from DNN 9.2.2

## Steps to reproduce
1. Exported a full site from DNN 9.2.2 on a local Windows 10 machine
2. Copied the export file to a fresh and empty DNN 9.3.1 site on my local machine and another empty site on a remotely hosted server
3. Selected Import Data from the Import/Export Persona Bar menu
4. The exported file shows up in the list of available imports
5. Hovering over the import file highlights it, but it cannot be selected (happened on both my local machine and a hosted server)

## Current result
Unable to select a highlighted import file in DNN 9.3.1

## Expected result
Should be able to select and process an import file

## Screenshots
See attached file
DNN_Import

## Error log
There was no response to the hover and select action. No error was created in the log. 

## Additional context
Tried to import one of several exported files.  None of them could be selected.

## Affected version
[x] 9.3.1
[x] 9.2.2

## Affected browser
Tried with the latest versions of these browsers.
 [x] Chrome
 [x] Edge

valadas:
Closes #718 

## Summary
This PR replaces the tooltip component for a normal html/css tooltip. The tooltip component was used here but modified to not show an icon and moved and restyled. The tooltip common component was recently modified to be accessible and was mainly tested in the context of a label tooltip.
So this solution has about the same footprint without "hacking" a component that was made for label tooltips.

image

### DIFF
--- a/Extensions/Settings/Dnn.PersonaBar.SiteImportExport/SiteImportExport.Web/src/components/ImportModal/PackageCardOverlay.jsx
+++ b/Extensions/Settings/Dnn.PersonaBar.SiteImportExport/SiteImportExport.Web/src/components/ImportModal/PackageCardOverlay.jsx
@@ -11,7 +11,7 @@ class PackageCardOverlay extends Component {
                 <div className="icon-container">
                     <div>{props.isSelected ? Localization.get("ClicktoDeselect") : Localization.get("ClicktoSelect")}</div>
                 </div>                
-            </div >
+            </div>
         );
     }
 }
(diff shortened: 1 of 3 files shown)


daguiler (approved):
Worked fine for me.

david-poindexter (approved):

PR #795, opened 2019-04-14
Fixed an issue with site map priority override

### LINKED ISSUE #223: Seo priority not saved right
## Description of bug
In Admin Settings, Seo, Sitemap settings, sitemap providers is the Override Priority not saved correctelly.
Dnn sitemap generation wait for a override value multiplied by 100 but the saved value is the real value (with decimal separator).

## Steps to reproduce
1. Go to 'Admin Settings, Seo, Sitemap settings, sitemap providers 
2. Save a value of 0.5 for the coreSitemapProvider
3. look in the db table portalsettings , record with coreSitemapProviderValue as SettingsName
4. the saved value is 0.5 in place of 50

## Current result
The priority field for all pages in sitemap.aspx equals 0

`<url>
<loc>https://www.vcomm.be/fr-fr/</loc>
<lastmod>2018-12-03</lastmod>
<changefreq>daily</changefreq>
<priority>0.0</priority>
</url>`

## Expected result
The priority field for all pages in sitemap.aspx equals 0.5
`<url>
<loc>https://www.vcomm.be/fr-fr/</loc>
<lastmod>2018-12-03</lastmod>
<changefreq>daily</changefreq>
<priority>0.5</priority>
</url>`

## Affected version

* [x] 9.2.2

valadas:
Closes #223 

## Summary
Right or wrong, the core site map provider is coded to store the value multiplied by 100 in order to prevent errors with decimal separators in different cultures. The Persona Bar SEO module was not accounting for that fact so the values need to be multiplied and divided by 100 to save the correct value into the database and return a correct value in the UI. This PR fixes that.

### DIFF
--- a/Extensions/Settings/Dnn.PersonaBar.Seo/Services/SeoController.cs
+++ b/Extensions/Settings/Dnn.PersonaBar.Seo/Services/SeoController.cs
@@ -478,7 +478,7 @@ namespace Dnn.PersonaBar.Seo.Services
                 {
                     p.Name,
                     p.Enabled,
-                    p.Priority,
+                    Priority = p.Priority / 100f,
                     p.OverridePriority
                 }).ToList();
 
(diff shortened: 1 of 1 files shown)


david-poindexter (approved):
Looks good to me - thanks @valadas!

donker (approved):

PR #693, opened 2019-04-09
Restructure AE Extensions into a single VS Project.

ohine:
## Summary

Currently this combines the all of the persona bar extensions into a new common project called `Dnn.PersonaBar.Extensions`. This reduces the projects in the solution to 

The goal for this phase is to end up with 3 DNN install files. Dnn.EditBar, Dnn.PersonaBar, and Dnn.PersonaBar.Extensions. Yarn workspaces will also be configured in the Dnn.PersonaBar.Extensions project to further improve our build times and developer productivity. 

## Testing Summary

Completed install with `v9.3.1.2` and [AE v3.0.0.8](https://dev.azure.com/dotnet/DNN/_build/results?buildId=13977). Before triggering the install, I deleted the `23` PersonaBar Extensions from ~/install/module and replaced them with the `21` from `v3.0.0.8`. No issues found.

Completed install with `v9.4.0.alpha-18` and [AE v3.0.0.10] Before triggering the install, I deleted the `23` PersonaBar Extensions from ~/install/module and replaced them with the `3` from `v3.0.0.10`. Users module not rendering due to missing export sub-bundle. Retesting with [v3.0.0.14](https://dev.azure.com/dotnet/DNN/_build/results?buildId=16019) and no issues detected. 

### Build Time Improvements
Full CI (clone, compile, yarn install/build, package, unit test)
*v2.0.0: 17min 30s
*v3.0.0: 11min 30s

### Reduction of Packages to Install/Upgrade
This reduces the packages installed and upgraded on platform releases by **50%** (v9.3.2 has 41 modules to install, v9.4.0 has 21). 

### Testing not yet completed.
Upgrade from `v9.3.1`, and `v9.2.2` to verify uninstallation of all old extensions and everything still functions under the new v3 package. 

## Areas for Future Improvements
`PersonaBar.UI` also contains a react.web project. This would benefit if it's loaded into the yarn workspaces configuration. The initial idea is to move the config for yarn workspaces up to the root. If that doesn't work, we will need to relocate `PersonaBar.UI` below the `WebApps` folder. The same issue will apply after `React.Common` is merged into the AE repo. 

Combine the UI and Library projects into a single project/assembly. `EditBar.UI` and `EditBar.Library` would become `EditBar`. `PersonaBar.UI` and `PersonaBar.Library` would become `PersonaBar`. This could potentially be a breaking change as any personabar extensions would need to be updated for the new reference.

PR #797, opened 2019-04-14
Uses source-map for webpack config

### LINKED ISSUE #419: Remove usage of inline-source-map
## Description of problem
Some modules use the inline-source-map devtool. This maps the transpiled minified javascript back to the original modern source code to make it easier to find where an error occurs in production. Using inline-source-map includes the whole transpiled + source code in the resulting bundle which makes is unnecessary huge.

## Description of solution
The source map should be in a separate file, browser devtools will use it only if open but the normal visitors will not load that map file. This would save bandwidth and probably improve performance. The only change needed is replacing inline-source-map for source-map and ensuring the .map file is included in the module packaging together with the bundle file.

## Screenshots
inline-source-map:
image

source-map
image

valadas:
Closes #419 

## Summary
Source maps are js files that help debug minified javascript by mapping the minified version to the original source code that generated them before transpiling. Webpack allows generating those files in multiple ways.

Some modules had the inline-source-map option which includes the source code inside the bundle file and makes it huge for no good reason. For those, this PR will reduce their footprint without loosing the features.

Some other modules had no source maps to begin with, this PR will make the cryptic error message more meaningful and help use pinpoint strange hard to reproduce issues and edge cases more easily. Note that the source map files are only loaded by the browser development tools, they do not affect normal site users in production.


ohine (changes requested):
This looks great! One minor tweak, can you easily fix the white space differences in the solution file? I’m worried about it conflicting with #693

valadas:
I am not sure what you mean @ohine if I view it (without ignore whitespace) I see no problem? Am I looking at it wrong, where do you see whitespace issues ?

donker (dismissed):

bdukes (approved):
I reverted the whitespace changes to the sln file (I saw it in some views but not others, I think the tools are just trying to be too helpful)

donker (approved):

ohine (approved):

valadas:
Thanks @bdukes can you explain me how you fixed this.

How you push a commit to someone else PR and how you revert a whole file commit ? Just want to learn this.

bdukes:
When you create a PR, there's a box checked by default, "Allow edits from maintainers"
image

So I was able to pull down your branch, add a commit, and push it back up to your branch.

For the whitespace change, I reverted your change to the `sln` file, and then used the [Commit View](https://git-extensions-documentation.readthedocs.io/en/latest/commit.html#undoing-or-resetting-changes) in [Git Extensions](https://gitextensions.github.io/) and (with the whitespace filter on so I only saw non-whitespace changes) reset the lines with substantial changes (that is, undid the revert, to leave them as they were).  What was then left was just the whitespace changes being reverted, so I committed what was left.

Does that help?

valadas:
That is kinda how I tried to do this in the past, but it never worked for me, not sure why. Maybe one day we can look at this together, see what I am doing wrong...

ohine:
This is how I've done it successfully when working with https://github.com/dnnsoftware/Dnn.AdminExperience/pull/717

```
git clone https://github.com/donker/Dnn.AdminExperience.git .
git remote add upstream https://github.com/dnnsoftware/AdminExperience.git
git fetch upstream
git checkout gdprv2
git merge upstream/development
```
Open Visual Studio, update references, build. 
```
git add ....
git commit -m "my message"
git push
```
I know @bdukes has a different workflow, which I wasn't able to get to work right (forgot the exact error). I'd love to figure out one of these days what I did wrong because his method seems much quicker/cleaner as it doesn't require cloning an entire fork I believe.

valadas:
Hmm, interesting. I think github instructions confused me, your way makes sense, you are actually allowed to push to the submitter branch then. I was trying to push to upstream or something... I will try that next time. Thanks.

bdukes:
Here are the commands I used:
```bash
git checkout release/2.0.x
git checkout -b valadas_issue-419 
git pull https://github.com/valadas/Dnn.AdminExperience issue-419
gitex
git push https://github.com/valadas/Dnn.AdminExperience valadas_issue-419:issue-419
```

I did not have a local copy of the `release/2.0.x` branch yet, so the first `checkout` created a local branch from `origin/release/2.0.x`. Otherwise, I would have needed to fetch and reset to make sure I was on the latest version of the destination branch.

`gitex` opened Git Extensions for the repo, and that's where I did the work to revert the `sln` file changes.

PR #798, opened 2019-04-15
Implement new data consent feature in Users PersonaBar extension

donker:
## Summary

In DNN 9.4.0 we're looking to implement a new feature called "Data Consent" which allows site administrators to require users to "sign" an agreement for the terms of use/privacy statement. 
In addition to signing an agreement (consenting to data usage), the user may also request removal from the site. This process can be done in various ways and one is "Manual" where the administrator will be required to do this him/herself. To that end the users module must give adequate options to manage this.
This PR aims to provide visual feedback to administrators as follows:

- The filter dropdown above the users table expanded with users that signed, those that didn't and thos that requested removal from the site.
- An icon to the left of the user showing the status of this user: signed, not signed or requests removal
- In the details screen of the user: has consented (i.e. signed), consented at what time/date and a red warning text if the user has requested removal

Note that all the above changes are invisible if the Data Consent feature is switched off.

Note 2: this also depends on the addition of a couple of SVG icons in React Common

Screenshot 2019-04-12 at 10 11 58

Screenshot 2019-04-15 at 09 48 28

Screenshot 2019-04-15 at 09 48 49

Screenshot 2019-04-15 at 09 49 26

Screenshot 2019-04-15 at 09 49 42

### DIFF
--- a/Extensions/Manage/Dnn.PersonaBar.Users/Components/Dto/UserFilters.cs
+++ b/Extensions/Manage/Dnn.PersonaBar.Users/Components/Dto/UserFilters.cs
@@ -7,6 +7,9 @@
         Deleted = 2,
         SuperUsers = 3,
         RegisteredUsers = 4,
-        All = 5
+        HasAgreedToTerms = 5,
+        HasNotAgreedToTerms = 6,
+        RequestedRemoval = 7,
+        All = 8
     }
 }
--- a/Extensions/Manage/Dnn.PersonaBar.Users/Components/UsersController.cs
+++ b/Extensions/Manage/Dnn.PersonaBar.Users/Components/UsersController.cs
@@ -72,7 +72,7 @@ namespace Dnn.PersonaBar.Users.Components
         public IEnumerable<KeyValuePair<string, int>> GetUserFilters(bool isSuperUser = false)
         {
             var userFilters = new List<KeyValuePair<string, int>>();
-            for (var i = 0; i < 6; i++)
+            for (var i = 0; i < 9; i++)
             {
                 userFilters.Add(
                     new KeyValuePair<string, int>(
@@ -81,6 +81,12 @@ namespace Dnn.PersonaBar.Users.Components
             if (!isSuperUser)
             {
                 userFilters.Remove(userFilters.FirstOrDefault(x => x.Value == Convert.ToInt32(UserFilters.SuperUsers)));
+            }
+            if (!PortalSettings.DataConsentActive)
+            {
+                userFilters.Remove(userFilters.FirstOrDefault(x => x.Value == Convert.ToInt32(UserFilters.HasAgreedToTerms)));
+                userFilters.Remove(userFilters.FirstOrDefault(x => x.Value == Convert.ToInt32(UserFilters.HasNotAgreedToTerms)));
+                userFilters.Remove(userFilters.FirstOrDefault(x => x.Value == Convert.ToInt32(UserFilters.RequestedRemoval)));
             }
             userFilters.Remove(userFilters.FirstOrDefault(x => x.Value == Convert.ToInt32(UserFilters.RegisteredUsers)));//Temporarily removed registered users.
             return userFilters;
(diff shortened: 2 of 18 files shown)


sleupold (changes requested):
IMO it doesn't make sense to rename column userPortals.authorised just in a few places.
Either we keep it or need to rename it in the table, view vw_users and all sprocs.
My suggestion would be keeping it as is and create a new issue for rename (if necessary).

donker:
> 
> 
> IMO it doesn't make sense to rename column userPortals.authorised just in a few places.
> Either we keep it or need to rename it in the table, view vw_users and all sprocs.
> My suggestion would be keeping it as is and create a new issue for rename (if necessary).

I think it is fine like this. The parameters is with a z and so is the column. This SPROC will one day be deprecated anyway when we begin to consilidate code. At that time we'll harmonize naming.

Extensions/Manage/Dnn.PersonaBar.Users/Components/Dto/UserFilters.cs
```diff
@@ -7,6 +7,9 @@ public enum UserFilters
         Deleted = 2,
         SuperUsers = 3,
         RegisteredUsers = 4,
-        All = 5
```
mtrutledge, line 10: Should the value of this Enum have changed to 8 from 5? Not sure if this is used anywhere else, but this seems like it may cause some issues or potential bugs. Would it have been better to leave All as 5 and added the new items after?

mtrutledge (commented):
Awaiting feedback on my comment for the UserFilter enum. The React side of things looks good.

Extensions/Manage/Dnn.PersonaBar.Users/Components/Dto/UserFilters.cs
donker: I thought about this and this was my reasoning:

1. when the filter is being requested through WebAPI it is serialised in order, so all would be somewhere in the middle

2. It would mean someone took a dependency *on the Users PersonaBar extension*. It's not a core enum. I don't know of anyone who would do something foolish like that. It would also surprise me that anyone who would live that dangerously would expect that to remain stable.

mtrutledge: That makes sense. I just didn't know if that would cause issues or not and thought I would ask.

Extensions/Manage/Dnn.PersonaBar.Users/Components/UsersController.cs
```diff
@@ -72,7 +72,7 @@ public IEnumerable<UserBasicDto> GetUsers(GetUsersContract usersContract, bool i
         public IEnumerable<KeyValuePair<string, int>> GetUserFilters(bool isSuperUser = false)
         {
             var userFilters = new List<KeyValuePair<string, int>>();
-            for (var i = 0; i < 6; i++)
+            for (var i = 0; i < 9; i++)
```
mitchelsellers: I know it was this way before, but in the interest of not breaking stuff in the future, can we get rid of the magic numbers here and have this switched to a foreach loop.  We are asking for trouble here with a hard-coded maximum value that is Max Enum value +1.

I know not your problem Peter, but trying to improve the quality as we go.  Sorry, I missed this on first review.

mitchelsellers (changes requested):

Extensions/Manage/Dnn.PersonaBar.Users/Components/UsersController.cs
donker: I can rework that. Give me a few mins.

donker: BTW I agree with you and thought the same but I felt "touch as little as possible"

donker: Check new code ...

mitchelsellers (approved):
Looks good!

ohine (approved):
These changes look great! I pushed a few minor tweaks to references and fixed a failing unit test to get the build to finish successfully in devops.

PR #906, opened 2019-05-02
Add "remove deleted users" button to place you'd expect it

donker:
Currently removing deleted users is handled by the "empty recycle bin" button on the recycle bin extension. This is (1) obscure and (2) not the place you'd expect to find it when managing users. On top of this you can't remove deleted users without also removing deleted pages etc. So there is also a loss of control.

In our efforts to "finish" the personabar I cannot see why we shouldn't have access to this function at the top of the users module.


ohine (approved):
Thanks for contribution! These changes look great.

mitchelsellers (approved):

PR #984, opened 2019-06-16
Fixes the Dnn 9.4.0 upgrade issue

### LINKED ISSUE #981: Persona Bar modules do not work on upgrade
## Description of bug
Almost all extensions are missing after upgrading. (except site assets and global assets)

## Steps to reproduce
1. Install Dnn 9.3.2
2. Upgrade to Dnn 9.4.0 alpha

## Current result
Most persona bar modules are missing

## Expected result
All modules should still exist

## Additional context
The upgrade should uninstall all persona bar modules and then reinstall them.

## Affected version

* [x] 9.4.0 only when upgrading

valadas:
## Summary
Closes #981 

This is a work in progress, the problem is that the IUpgradeable code runs only after the installation is finished and the site reloads. Because it was uninstalling the modules, on upgrades we effectivelly install them then right after uninstall them. 

The reason we where doing this was in order to unregister and delete the old PB assemblies (that are now merged into a single assembly). 

One possible solution is that instead of uninstalling the modules, we could just unregister their old assemblies and delete them.

This is a work in progress for this solution, it currently unregisteres the assemblies, but there is an issue deleting the files. I put it here in case someone has time to check it out since I gotta leave for the evening. I _think_ it might just be the starting \\ in the file path to delete but have not verified this theory yet...


mitchelsellers (dismissed):

Library/Dnn.PersonaBar.UI/Dnn.PersonaBar.UI.dnn
```diff
@@ -150,7 +150,17 @@
                         </file>
                     </files>
                 </component>
-                <component type="Cleanup" version="01.03.00" fileName ="01.03.00.txt" />
+                <component type="Cleanup" version="03.00.00" fileName ="03.00.00.txt" />
```
ohine: It appears v1.3.0 cleanup was removed and v3.0.0 is included twice.

ohine (changes requested):

mitchelsellers (approved):

ohine (approved):

PR #985, opened 2019-06-17
Fixes css console error on PB modules

### LINKED ISSUE #974: Persona bar modules with no css files cause browser error
## Description of bug
There was a cleanup done on the empty css files that many persona bar modules, since then, they still try to load but do not exist so give a 404 error in the browser console.
Additionally, the request url has a incorrect function call in the actual url.

## Steps to reproduce
1. Go to a couple of persona bar modules with browser console open
2. Notice errors

## Current result
404 error and wrong parameters in querystring

## Expected result
Ideally should not load missing files, as the second best we need to restore the empty css files to prevent the 404 errors and fix the querystring.

## Screenshots
image

## Affected version

* [x] 9.4.0
* [ ] 9.3.2
* [ ] 9.3.1
* [ ] 9.2.2
* [ ] 9.2.1
* [ ] 9.2
* [ ] 9.1.1
* [ ] 9.1
* [ ] 9.0

valadas:
Closes #974 

Utils.js code used to use urlArgs (provided by require.js) to append cache-busting to css files in debug mode.

Starting at some version requireJs supports either a string or a function but before it used to support only a string. So this PR first checks if urlArgs is a string, if so does the previous behavior, if not just ignores it. This is in order to try to minimize the risk if module developers have used strings here for requrests in the init function of persona bar modules.

It _should_ be no risk for third party modules using this as a string. If module developers decide to use it as a function then they can manage their own injection using that function would would not need Dnn to do it for them.

### DIFF
--- a/EditBar/Dnn.EditBar.UI/editBar/scripts/util.js
+++ b/EditBar/Dnn.EditBar.UI/editBar/scripts/util.js
@@ -166,7 +166,11 @@ define('css',{
             path = config.baseUrl + path;
         }
 
-		inject(path + '?' + config.urlArgs);
+        if (typeof config.urlArgs === 'string') {
+            path = path + '?' + config.urlArgs;
+        }
+        inject(path);
+
 		load(true);
 	},
 	pluginBuilder: './css-build'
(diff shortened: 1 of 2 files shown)


mitchelsellers (approved):

bdukes (approved):

ohine (approved):
Thanks for getting this fixed @valadas! Currently waiting for the devops builds to finish before merging.

PR #1031, opened 2019-07-26
() Characters that are not illegal are considered illegal characters

### LINKED ISSUE #1029: () Characters that are not illegal are considered illegal characters
## Description of the bug
When a user tries to add S.E.O new URL with special characters **(** ,or **)** it'll not work while the error message doesn't really say that these characters are illegal characters as you can see in this [recording](https://www.loom.com/share/2c2049609f314b198000166e877b4247)

## Steps to reproduce
1. Go to any page's settings 
2. SEO Tab 
3. Add New URL > type in a URL like (test)
4. Click Save

## Current result
An error message that illegal characters are in the URL is shown.

## Expected result
An error message should not be shown or fix the error message.

## Screenshots
image

## Affected version

* [x] 9.3.2
* [x] 9.3.1
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

aelibyary:
Fixes #1029 

## Summary
This issue is very obscure. The error message is misleading.

The URL is rejected because of a hidden host setting named AUM_ReplaceChars which has a default value of
 &$+,/?~#<>()¿¡«»!"
(including a space at the beginning)

URLs are tested against the above characters. In a nutshell, the above characters are replaced with a hyphen when found in the URL.

The other aspect of this issue is that the validation logic returns a suggested modified URL to the front-end, which is completely ignored.

@daguiler Suggested:

I think the easiest way to fix this without introducing breaking changes is
1- to change the error message so that it doesn't specify which characters are "illegal", (**Fixed**)
2- and also, replace the user-provided URL with the suggested modified URL in case there were validation errors. (**Fixed**)

[Fix Video](https://drive.google.com/open?id=1_VUV9xKx4sv7ZWCWKYQ7FJKl7Z8TrC_e)

### DIFF
--- a/Extensions/Content/Dnn.PersonaBar.Extensions/WebApps/Pages.Web/src/actions/pageSeoActions.js
+++ b/Extensions/Content/Dnn.PersonaBar.Extensions/WebApps/Pages.Web/src/actions/pageSeoActions.js
@@ -60,6 +60,18 @@ const pageSeoActions = {
                         data: { error: response.ErrorMessage }
                     });
                     utils.notifyError(response.ErrorMessage);
+
+                    // DNN-30998: Adding SuggestedUrl returned by the server in case of validation errors
+                    if(response.SuggestedUrlPath)
+                    {
+                        dispatch({
+                            type: ActionTypes.SEO_CHANGE_URL,
+                            payload: {
+                                key: "path",
+                                value: response.SuggestedUrlPath
+                            }
+                        });
+                    }
                     return;
                 }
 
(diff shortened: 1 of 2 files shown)


bdukes (approved):

valadas (approved):

v-karbovnichy:
@bdukes @valadas @mitchelsellers @daguiler can someone merge this, it has 2 approvals already.

valadas:
As far as I know we are waiting to release 9.4.0 before creating the 9.4.1 branch and then we will start merging, correct ?

v-karbovnichy:
So this is the expected events timeline:
* make release 9.4.0 from `release/9.4.x`
* merge this PR into `release/9.4.x`
* make release 9.4.1 from `release/9.4.x`

correct?

mitchelsellers:
Correct

valadas:
It is my understanding too

valadas:
@aelibyary this is currently targetting development which means it would be included in Dnn 10, would you like to re-target the release/3.0.x branch for inclusion in Dnn 9.4.1 ?

aelibyary:
> @aelibyary this is currently targetting development which means it would be included in Dnn 10, would you like to re-target the release/3.0.x branch for inclusion in Dnn 9.4.1 ?

Hello @valadas , #1166 has been created to target release/3.0.x branch

PR #1032, opened 2019-07-26
Fixed typo in cleanup script

### LINKED ISSUE #1027: Dnn.PersonaBar.Pages.dll is not cleaned up during upgrade
The Persona Bar dlls where merged into one and the old dlls are supposed to be removed. One exception is the Dnn.PersonaBar.Pages.dll which stays behind and may cause duplicated code or some unforeseen issues.

valadas:
Fixes a typo in the cleanup script, 

Should close #1027 and probably fixes https://github.com/dnnsoftware/Dnn.Platform/issues/2902


mitchelsellers (approved):

david-poindexter (approved):
Good catch - thanks @valadas

PR #1033, opened 2019-07-26
Adds back empty Roles.css file

### LINKED ISSUE #1028: Dnn.Roles/css/Roles.css only exists on upgrade
The file in the title exists only when upgrading, on a clean install it does not exist on the filesystem. This was part of a fix about 404's in the browser console and we previously decided to bring back those empty css files as a quick fix. See #1025

valadas:
Adds back the empty Roles.css file in order to close #1028 
this prevents 404 errors in browser console


mitchelsellers (approved):

david-poindexter (approved):
Thanks @valadas

Issue #1071, opened 2019-07-31
Cannot search delete users


valadas:
## Description of bug
In the users module when viewing deleted users, search does not work and returns nothing.

## Steps to reproduce
List the steps to reproduce the behavior:
1. Open the users module
2. Set the show dropdown to Deleted Users
3. Try to search for a user that exists

## Current result
No users returned

## Expected result
Should be able to return results

## Error log
No errors logged

## Additional context
This was found after investigating possible side-effects of #1065 when privacy settings are on. When on, those settings add some items in this dropdown and we where worried that search may not work for some types. Ends up that all the new types work perfectly but deleted users.

I _think_ but have not investigated a lot yet, that the problem is that the stored procedure for the search checks for IsDeleted in the Users table instead of the PortalUsers table.

Which brings me to the question of why is this duplicated, do we need IsDeleted in both tables? How does this duplication affect hard deletes, and how does it affect the new privacy settings for users that request removal with mutiple possible removal options.

@donker if you have any more insight on this, please add comments here.

All this being said, #1065 is already merged and I think it brings us in a better situation than before, just loggin this issue so we can fix the last type for deleted users at some point.

## Affected version

* [x] 9.4.0
* [x] 9.3.2 ( it was on more types )

PR #1072, opened 2019-07-31
Fix deleted users search

donker:
The SPROC for searching users included deleted user checks which bound to the users table instead of the PortalUsers table where it should look when searching regular users.


valadas (approved):
Tested and it works as expected, with this change we can now search for any types

valadas:
Closes #1071

david-poindexter (approved):

PR #1073, opened 2019-07-31
Link to DNNdocs

Tychodewaard:
I think the link to documentation should be a link to DNNDocs

### DIFF
--- a/Extensions/Content/Dnn.PersonaBar.Extensions/WebApps/Licensing.Web/src/components/platform/index.jsx
+++ b/Extensions/Content/Dnn.PersonaBar.Extensions/WebApps/Licensing.Web/src/components/platform/index.jsx
@@ -39,7 +39,7 @@ class Platform extends Component {
     }
 
     onDocCenterClick() {
-        window.open("http://www.dnnsoftware.com/docs/", "_blank");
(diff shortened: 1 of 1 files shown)


bdukes (approved):

mitchelsellers (approved):

valadas (approved):

valadas:
@Tychodewaard  this is currently targetting development which means it would be included in Dnn 10, would you like to re-target the release/3.0.x branch for inclusion in Dnn 9.4.1 ?

Tychodewaard:
@valadas I really want to but how do I do that? BTW: I would expect that I am not in charge of deciding whether it should be in what release. That would be too much honour :-) 

I can see you added it to 9.4.1. on August 1 and yesterday to milestone 9.4.1 and 10.0.0.
Is there something else needed or to be done on my side?

bdukes:
It looks like this change was incorporated into a previous commit already: 147748a2889f67e8690d712cdb8bd39385794a27

We'll go ahead and close this one, but thank you so much for starting the contribution, we really do appreciate it.

david-poindexter:
Yes, I made this change a while back and it is already in 9.4.0.  Thanks.

PR #1089, opened 2019-08-07
Blank_Name_is_accepted_as_Group_name

SergeyDryomin:
Fixed "Blank name is accepted as Group name"
Added validation for white spaces string, when new group added.
*No UT added - to add it have to add UT new project
## Summary
https://github.com/dnnsoftware/Dnn.Platform/issues/2927
DNN-18333
Note: same issues possible to fix into 'DNN Platform\Library\Common\Requires.cs' (with adding new method or just to change NotNullOrEmpty to IsNullOrWhiteSpace validation)

### DIFF
--- a/Extensions/Content/Dnn.PersonaBar.Extensions/Services/RolesController.cs
+++ b/Extensions/Content/Dnn.PersonaBar.Extensions/Services/RolesController.cs
@@ -401,13 +401,13 @@ namespace Dnn.PersonaBar.Roles.Services
 
         private void Validate(RoleGroupDto role)
         {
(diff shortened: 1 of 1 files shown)


valadas (approved):
Looks good to me, if you have time to do the other PR for the Platform, it would be even nicer :)

bdukes (approved):

valadas:
@sergeydryomin  this is currently targetting development which means it would be included in Dnn 10, would you like to re-target the release/3.0.x branch for inclusion in Dnn 9.4.1 ?

PR #1093, opened 2019-08-08
DNN-28934: Fix Export to PDF Issue on SQL Console Issue

### LINKED ISSUE #1092: Export to PDF Issue on SQL Console
## Description of bug
When a query result on the SQL Console has long text cells, which are wrapped in the result table; then the first exported PDF is normal but next exported PDF files are distorted as wrapped texts are not wrapped anymore on PDF files.

## Steps to reproduce
1. Log in as a Super User
2. Go to Persona Bar > Settings > SQL Console
3. Write a query like `SELECT * FROM Tabs` and click the "Run Script" button
4. Observe that the result table is generated and some columns like IconFile, VersionGuid, UniqueId and etc. have wrapped long texts.
5. Click on Export > Export to PDF and wait for the PDF to be generated and downloaded.
6. Open the PDF file and see that it is normal; all long texts are wrapped as expected.
7. Click on Export > Export to PDF and wait for the PDF to be generated and downloaded for a second time.
8. Open the PDF file and see that, this time long texts are not wrapped and the PDF layout is distorted.

## Current result
PDF files containing long text values in any table cell are distorted after 2nd export.

## Expected result
PDF files should be consistently generated as the same without any distortion.

### 1st Export
image

### 2nd Export
image

## Error log
N/A

## Additional context
N/A

## Affected version
* [x] 9.3.2
* [x] 9.3.1
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser
* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

### LINKED ISSUE #3116: Export to PDF Issue on SQL Console
## Description of bug
When a query result on the SQL Console has long text cells, which are wrapped in the result table; then the first exported PDF is normal but next exported PDF files are distorted as wrapped texts are not wrapped anymore on PDF files.

## Steps to reproduce
1. Log in as a Super User
2. Go to Persona Bar > Settings > SQL Console
3. Write a query like `SELECT * FROM Tabs` and click the "Run Script" button
4. Observe that the result table is generated and some columns like IconFile, VersionGuid, UniqueId and etc. have wrapped long texts.
5. Click on Export > Export to PDF and wait for the PDF to be generated and downloaded.
6. Open the PDF file and see that it is normal; all long texts are wrapped as expected.
7. Click on Export > Export to PDF and wait for the PDF to be generated and downloaded for a second time.
8. Open the PDF file and see that, this time long texts are not wrapped and the PDF layout is distorted.

## Current result
PDF files containing long text values in any table cell are distorted after 2nd export.

## Expected result
PDF files should be consistently generated as the same without any distortion.

### 1st Export
image

### 2nd Export
image

## Error log
N/A

## Additional context
Note that non-wrapped texts do not have any separators or spaces.

## Affected version
* [x] 9.3.2
* [x] 9.3.1
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser
* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

kkadir:
Fixes #1092 

## Summary
The word-based encoding/rendering option `getWords(characters)` is removed as it should always be based on characters only to wrap long texts without separators and/or spaces.

**[Confirmation Video](https://drive.google.com/open?id=1tEHOU16bYb7w_i47z_MuhqdwInYjSDB8)**

### DIFF
--- a/Extensions/Content/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.SqlConsole/scripts/jspdf.js
+++ b/Extensions/Content/Dnn.PersonaBar.Extensions/admin/personaBar/Dnn.SqlConsole/scripts/jspdf.js
@@ -13446,9 +13446,10 @@ Q\n";
     NodeParser.prototype.paintText = function(container) {
         container.applyTextTransform();
         var characters = window.html2canvas.punycode.ucs2.decode(container.node.data);
-        var textList = (!this.options.letterRendering || noLetterSpacing(container)) && !hasUnicode(container.node.data) ? getWords(characters) : characters.map(function(character) {
-            return window.html2canvas.punycode.ucs2.encode([character]);
-        });
+        
+		var textList = characters.map(function(character) {
+			return window.html2canvas.punycode.ucs2.encode([character]);
+		});
 
         var weight = container.parent.fontWeight();
         var size = container.parent.css('fontSize');


bdukes (changes requested):
`jspdf.js` is a third party script (https://github.com/MrRio/jsPDF), we shouldn't be making changes to it (we probably shouldn't be checking it in, in the first place, but that's a separate discussion).  You can see if there's an option you can pass to it which achieves your desired result, or an update to the library, but making a change like this in the library is going to get us in trouble down the road when we need to update it.

valadas (changes requested):
I agree, modifying a third party library is not a great idea for the future maintenance.

valadas:
Taking a very quick looks at their current version, I do not even see that method in the code, so I guess there was some major updates in later releases. Maybe upgrading the library would be a better solution to this problem. @kkadir I am closing this PR but keeping the issue open if you have time to try out the new version, it would be a better PR.

kkadir:
Thank you for your updates @bdukes and @valadas.

But the 3rd-party jsPDF was not included as a library nor a package but instead included into the module as a *.js file. That is why I did not consider it as a 3rd-party library update.

Please also note that the current update will only affect SQL Console as the updated jsPDF belongs only to that module and other modules have their own jsPDF.js files. So it will be great if you can reconsider this PR, thanks in advance!

PR #1098, opened 2019-08-10
Fixes an issue when creating a new site with a new admin

### LINKED ISSUE #1096: "passwords do not match" error when creating new portals
When trying to create a new portal, I get this error.  The workaround is easy: just turn off "Assign Current User as Administrator", then erase the password in the password field. 
I believe the issue is with chrome saving usernames and passwords, then automatically putting them in fields, even if they are hidden.

valadas:
Closes #1096 

## Summary
* Fixes a css issue that prevented the Collapsible panel to properly calculate the form height and was constantly 0px high, so not visible when creating a new site without using the current user as the new admin
* Workaround to prevent Chrome for autofilling the new user fields from the current user saved credentials.


mitchelsellers (approved):
LGTM @dnnsoftware/tag can I get a second review & merge on this ASAP for 9.4.0

bdukes (approved):

PR #1101, opened 2019-08-14
Build front end projects in parallel using lerna

### LINKED ISSUE #1100: Add Lerna to build in parallel
Lerna is very popular to manage mono-repos, in our case we could start by using it to build our front-end in parallel to speed up the build.

valadas:
#closes #1100 

Adds lerna in order to build frontend project in parallel while still using yarn workspaces, should significantly improve build times...

### DIFF
--- a/Build-Yarn-Workspace.ps1
+++ b/Build-Yarn-Workspace.ps1
@@ -1,3 +1,3 @@
 
 yarn install
-yarn workspaces run build
+yarn lerna run build --parallel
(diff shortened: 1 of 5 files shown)


valadas:
I am a bit disapointed, it save more than 2 minutes locally but in CI it had almost no effect. I will try to re-run in in CI at a couple different times just to rule out devops being slow right now...

valadas:
I think it boils down to the CI VM being only 2 vCPU... So unfortunately this does not affect much CI but I think it is still a great improvement for local builds.

mitchelsellers (approved):

david-poindexter (approved):

PR #1117, opened 2019-08-16
DNN-29164 - Open page settings can't display data for correct page

### LINKED ISSUE #1116: Open page settings can't display data for correct page
## Description of bug
Open page settings can't display data for the correct page. This error happens intermittently. Once the cache (both browser and the server) is loaded, this error doesn't happen most of the time so that IIS reset and browser cache reset is required in order to be able to reproduce this issue. Even after cache resets, multiple runs might be required since this bug occurs when a specific XHR request responds before another one.

## Steps to reproduce
Following reproduction steps focus on Google Chrome, steps (related to cache resets) might differ for other browsers:
1. Install 9.3.0 or use 9.4.x latest build and login as host; 
2. Open Settings > Site Settings > Languages in PB, add a new language, enable content localization; 
3. Restart IIS
4. Open Developer Tools from Chrome by clicking F12
5. Right-click on the browser refresh icon.
6. Choose "Empty Cache and Hard Reload"
7. Open/Refresh Languages again (same as Step 2)
8. Edit the pages for new language, click "view page settings" for any page except home page;
9. If not reproduced, go to Step 2.

In case it's not reproduced even after multiple runs, let me know and I'll shoot a reproduction video.

## Current result
The page details form will show data for home page, but not the selected page.

## Expected result
The page details form should show data for selected page.

## Screenshots
The following video misses the IIS reset and browser cache reset steps but just to give you an idea: https://www.screencast.com/t/nl2BPkmmb

## Affected version

* [x] 9.4.x
* [x] 9.3.2

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

berkarslan-xo:
Fixes #1116
## Summary

When the script got loaded the first time, a wrong page id was getting sent in the XHR request which caused this issue (there were multiple page requests being sent and the order of responses for that requests play an important role in this bug's intermittent occurrence). 

Demo: https://drive.google.com/open?id=1RvwR04B8iPrDdqrMAInCK3N0kN1cWMe2

### DIFF
--- a/Extensions/Content/Dnn.PersonaBar.Extensions/WebApps/Pages.Web/src/components/App.jsx
+++ b/Extensions/Content/Dnn.PersonaBar.Extensions/WebApps/Pages.Web/src/components/App.jsx
@@ -93,7 +93,7 @@ class App extends Component {
         window.dnn.utility.setConfirmationDialogPosition();
         window.dnn.utility.closeSocialTasks();
         this.props.getPageList().then(() => {
-            const selectedPageId = utils.getCurrentPageId();
+            const selectedPageId = utils.getSelectedPageId() || utils.getCurrentPageId();
             selectedPageId && !utils.getIsAdminHostSystemPage() && this.onLoadPage(selectedPageId);
         
             if (viewName === "edit") {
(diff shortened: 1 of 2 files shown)


bdukes (approved):

v-karbovnichy (approved):

valadas (approved):
Looks good

PR #1126, opened 2019-08-23
DNN-30135 - Issue when loading the Persona Bar: Not Thread Safe

### LINKED ISSUE #1125: Issue when loading the Persona Bar: Not Thread Safe
## Description of bug
An error, which has the screenshot of it below, is present sometimes several times per day.

## Steps to reproduce
Customer has the error frequently; however, unable to reproduce the error on demand, it just happens randomly. 

## Current result
The following error appears:
```
2019-04-01 10:04:31,130 [JWWEBSVR21][Thread:44][FATAL] DotNetNuke.Framework.PageBase - An error has occurred while loading page. 
System.InvalidOperationException: Collection was modified; enumeration operation may not execute. 
at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource) 
at System.Collections.Generic.List`1.Enumerator.MoveNextRare() 
at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source, Func`2 predicate) 
at Evoq.PersonaBar.UI.Controllers.EvoqPersonaBarContainer.Initialize(UserControl personaBarControl) 
at Dnn.PersonaBar.UI.UserControls.PersonaBarContainer.OnInit(EventArgs e) 
at System.Web.UI.Control.InitRecursive(Control namingContainer) 
at System.Web.UI.Control.AddedControl(Control control, Int32 index) 
at DotNetNuke.UI.Skins.Skin.InjectControlPanel() 
at DotNetNuke.UI.Skins.Skin.OnInit(EventArgs e) 
at System.Web.UI.Control.InitRecursive(Control namingContainer) 
at System.Web.UI.Control.AddedControl(Control control, Int32 index) 
at DotNetNuke.Framework.DefaultPage.OnInit(EventArgs e) 
at System.Web.UI.Control.InitRecursive(Control namingContainer) 
at System.Web.UI.Page.ProcessRequestMain(Boolean includeStagesBeforeAsyncPoint, Boolean includeStagesAfterAsyncPoint) 
```

## Expected result
This error shouldn't occur.

## Screenshots
image

## Affected version

* [x] 9.4.x

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

berkarslan-xo:
Fixes #1125 
## Summary

The problem lies in "Dnn.PersonaBar.Library.Repository.PersonaBarRepository.GetMenuItem(string identifier)" method which is called from "Evoq.PersonaBar.UI.Controllers.EvoqPersonaBarContainer.Initialize(UserControl personaBarControl)". GetMenuItem method doesn't exist in the StackTrace because JIT (Just-In-Time) Compiler does Inline Code Optimizations for release builds which basically means that some methods (especially like the ones that are only a few lines) get inlined into the previous method chain for performance reasons. Therefore, if you have a method call chain like A -> B -> C, it might become A -> B if C is inlined (you might think of it as method C's code gets copied under method B). 

Using a list instance which might change state in loops (FirstOrDefault method iterates the elements one by one which behaves like a loop) will raise that error when the list gets updated while iterating. The solution is to use ToList method to copy the list's current state into another list and then use it instead for iteration. You can watch the following video I shot which demonstrates the issue: https://drive.google.com/open?id=13bneeDRu2aqicSvJ-QtV2okoPUivkHh4

### DIFF
--- a/Library/Dnn.PersonaBar.Library/Repository/PersonaBarRepository.cs
+++ b/Library/Dnn.PersonaBar.Library/Repository/PersonaBarRepository.cs
@@ -67,12 +67,12 @@ namespace Dnn.PersonaBar.Library.Repository
 
         public MenuItem GetMenuItem(string identifier)
         {
-            return GetMenu().AllItems.FirstOrDefault(m => m.Identifier.Equals(identifier, StringComparison.InvariantCultureIgnoreCase));
+            return GetMenu().AllItems.ToList().FirstOrDefault(m => m.Identifier.Equals(identifier, StringComparison.InvariantCultureIgnoreCase));
         }
 
         public MenuItem GetMenuItem(int menuId)
         {
-            return GetMenu().AllItems.FirstOrDefault(m => m.MenuId == menuId);
+            return GetMenu().AllItems.ToList().FirstOrDefault(m => m.MenuId == menuId);
         }
 
         public void SaveMenuItem(MenuItem item)


v-karbovnichy:
@dnnsoftware/tag
Can someone review this please?

Library/Dnn.PersonaBar.Library/Repository/PersonaBarRepository.cs
```diff
@@ -67,12 +67,12 @@ public PersonaBarMenu GetMenu()
 
         public MenuItem GetMenuItem(string identifier)
         {
-            return GetMenu().AllItems.FirstOrDefault(m => m.Identifier.Equals(identifier, StringComparison.InvariantCultureIgnoreCase));
+            return GetMenu().AllItems.ToList().FirstOrDefault(m => m.Identifier.Equals(identifier, StringComparison.InvariantCultureIgnoreCase));
```
mitchelsellers, line 70: We need to get a bit more information on what this is trying to fix.  As the usage of FirstOrDefault is recommended by Microsoft, Resharper, and well pretty much any tool when the only call is to "FirstOrDefault".  

Is this possibly masking a different issue?

mitchelsellers (changes requested):

Library/Dnn.PersonaBar.Library/Repository/PersonaBarRepository.cs
berkarslan-xo: I changed the video's permission so that you should be able to watch it now. I have described the behavior in that video. 

To summarize the video, since "AllItems" property references a cache instance, I put ToList method to create a temporary copy of it and apply FirstOrDefault in that temporary variable. Otherwise, FirstOrDefault was working on cache instance's itself which was not safe and might throw this error when cache gets updated.

mitchelsellers: Looking at the video, I'm still not sure that we have an apples to apples comparison here.

The risk of doing this, is that we are going to be copying this collection around a ton, which has memory and other impacts to the downstream performance of the application.

berkarslan-xo: Actually, it will only do a shallow copy which means that only the references will get copied and not the values. The references will still point to the same values so that we should be fine as for performance reasons too.

v-karbovnichy: @mitchelsellers I agree with the solution here.

mitchelsellers: This is good enough for now

mitchelsellers (approved):

valadas (approved):
Looks ok to me

PR #1132, opened 2019-08-29
remove the dependency of security analyzer on evoq's default.aspx

### LINKED ISSUE #1131: remove the depdendency of security analyzer on evoq's default.aspx
## Description of bug
remove the depdendency of security analyzer on evoq's default.aspx

zyhfish:
Fix #1131


Extensions/Content/Dnn.PersonaBar.Extensions/Components/Security/Utility.cs
```diff
@@ -185,23 +186,8 @@ public static string GetFileCheckSum(string fileName)
 
         public static string GetApplicationVersion()
         {
-            return DotNetNukeContext.Current.Application.Version.ToString(3);
```
bdukes, line 188: Is there an issue using `DotNetNukeContext.Current.Application.Version`?

Extensions/Content/Dnn.PersonaBar.Extensions/Components/Security/Checks/CheckDefaultPage.cs
```diff
@@ -59,9 +59,8 @@ private bool CheckDefaultPageModified(out IList<string> modifiedFiles)
             var sumData = Utility.LoadFileSumData();
 
             var appVersion = Utility.GetApplicationVersion();
-            var appType = Utility.GetApplicationType();
 
-            var dataNodes = sumData.SelectNodes("/checksums/sum[@version=\"" + appVersion + "\"][@type=\"" + appType + "\"]");
+            var dataNodes = sumData.SelectNodes("/checksums/sum[@version=\"" + appVersion + "\"]");
```
bdukes, line 63: From a security perspective, do we care which version the page matches?  So long as it's a match, it should be valid, right?

bdukes: I recognize that's a separate change, but worth discussing while we're in here.

mitchelsellers: I don't believe we care, the general concept here is that it needs to be in the list of versions.  There are a few thoughts that I have about this, but we can take that offline if we want to dive in more.

Extensions/Content/Dnn.PersonaBar.Extensions/Components/Security/Utility.cs
zyhfish: HI @bdukes , this because Evoq and Platform checked application version with different ways, and will get different versions in some releases, so here need to change to always get the version from platform assembly.

Extensions/Content/Dnn.PersonaBar.Extensions/Components/Security/Checks/CheckDefaultPage.cs
zyhfish: Here I suggest to have the version match, it can help to verify the files are matched to the release.

bdukes (approved):

valadas (approved):
Looks good

PR #1137, opened 2019-09-02
Unable to set membership settings "Use Email Address as Username" ON

### LINKED ISSUE #1135: Unable to set membership settings "Use Email Address as Username" ON
## Description of bug
For some reason enabling "Use Email Address as Username" is returning an error message "Could not update settings"
See the
[recording](https://drive.google.com/file/d/1L9ob3XsCVue3zvje1wv2DBmbXNo2wmJO/view)

## Steps to reproduce
1. Navigate to Settings > Security page 
2. Click on Member Accounts tab 
3. Click on Registration Settings tab and select "Use Email Address as Username" toggle button on 
4. Click on Save button

## Current result
It is showing error message "Could not update settings"

## Expected result
It should save the Registration settings and return a success message.

## Screenshots
screenshot-1

## Additional context
This issue is not always reproducible but it happens mainly after adding couple of users.

## Affected version
* [x] 9.3.2
* [x] 9.3.1
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser
* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

aelibyary:
Fixes #1135

## Summary
This issue is very obscure. The error message is misleading.

RC:
The saving of the settings is being rejected because of a data issue not a code issue,
So to be able to reproduce this issue, you've first to create 2 users with the same Email Address, and as long as the system allows something like that, you won't probably expect that this might be the root cause of this settings problem,

Current behavior: 
From UI perspective we actually get a response from the server saying what exactly is the issue, however we ignores this error and simply show a default error message "Could not update the registration settings. Please try later."

Proposed solution:
Showing an error tool tip beside the failing setting displaying exactly the validation error occured (the response from the server)

Confirmation Video:
https://drive.google.com/file/d/17Dfs_JYOfrxvw2T-4wG4ipNWm-Idwjgz/view


daguiler (changes requested):
Hi @aelibyary, this looks good to me. However, it's targetting v10.
Can you close this PR, rebase your changes off `release/3.0.x` and re-submit please?

Are you ok with that @mitchelsellers?

Thanks!

aelibyary:
PR #1143 has been created as per @daguiler request above.

PR #1139, opened 2019-09-03
Search crawlers fails when running on different instances of web farm

### LINKED ISSUE #2961: Search Crawlers on different web servers
## Description of Problem
There are few issues reported to us related to search functionality. After troubleshooting it was detected that search crawlers are failing due to corrupted/missing files in a `App_Data\Search` folder. The errors are happening for the Evoq customers that are on web farm infrastructure setup. Apart of `Site Crawler`, there are also `File Crawler` and `Url Crawler` configured that uses lucene search engine API implemented on a Platform. It happened, that customer has crawlers configured to run on a different web servers. Because of that, files are getting modified and attempted to be scanned by different machines. It causes errors and search doesn't work as expected. 
This can be fixed by configuring all crawlers to be run on a same web server. 
Apart of that, to avoid same problems in a newer version, it is recommended to prevent admins to setup different machines for search crawlers. 

## Proposed Solution
Solution which is proposed here is to warn admin when he is attempting to setup crawlers like mentioned above. For that, it would be good to have a Host/Portal setting `SchedulersToRunOnSameWebServer` with a list of schedulers that we recommend to run on a same machine. On a scheduler configuration save, we will be checking if server that is going to be saved matches to the server of other schedulers in the list. If not, warning message will be displayed. This should not block admin to save it in other way, but supposed to notify only.
To use this feature, we need this validation to be implemented in a core part. 

Dear community, please let me know whether proposed solution makes sense and looks good to you to be implemented here?

## Affected version

* [*] 10.0.0 nightly build
* [*] 9.4.1 nightly build
* [*] 9.4.0 release candidate
* [*] 9.3.2 latest supported release

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.Platform/issues/2961

Issue fixed by checking whether `SchedulersToRunOnSameWebServer` host setting exists and creating list of recommended servers to be provided to client side. 
When saving Scheduler settings on UI, system validates whether entered Server matches to the servers configured for other schedulers in that list.

See demo: [DEMO](https://drive.google.com/file/d/1c716ocdbknBIaEwoy-Qv8wuMplebCqQO/view?usp=sharing)

### DIFF
--- a/Extensions/Content/Dnn.PersonaBar.Extensions/Components/TaskScheduler/TaskSchedulerController.cs
+++ b/Extensions/Content/Dnn.PersonaBar.Extensions/Components/TaskScheduler/TaskSchedulerController.cs
@@ -30,8 +30,8 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
-using System.Net;
 using DotNetNuke.Common.Utilities;
+using DotNetNuke.Entities.Controllers;
 using DotNetNuke.Instrumentation;
 using DotNetNuke.Services.Localization;
 using DotNetNuke.Services.Scheduling;
@@ -136,6 +136,7 @@ namespace Dnn.PersonaBar.TaskScheduler.Components
             scheduleItem.Servers = string.IsNullOrEmpty(servers) ? Null.NullString : servers;
             return scheduleItem;
         }
+
         public IEnumerable<ScheduleItem> GetScheduleItems(bool? enabled, string serverName = "", string taskName = "")
         {
             try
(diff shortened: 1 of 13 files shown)


valadas (approved):
Looks good to me

mitchelsellers (approved):

Extensions/Content/Dnn.PersonaBar.Extensions/Components/TaskScheduler/TaskSchedulerController.cs
```diff
@@ -167,5 +168,43 @@ public IEnumerable<ScheduleItem> GetScheduleItems(bool? enabled, string serverNa
...
+                if (!int.TryParse(scheduler.Trim(), out int id) || id == schedulerId)
+                {
+                    continue;
+                }
+
+                var schedule = SchedulingProvider.Instance().GetSchedule(id);
+
+                if (schedule != null && schedule.Enabled && !string.IsNullOrWhiteSpace(schedule.Servers))
+                {
+                    recommendedServers.AddRange(schedule.Servers.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries));
```
Yabgu, line 203: You are not trimming before distinct! If your server list holds values like in the below you may run into trouble:
`[  "1  ,   2,   x,y   \r\n....",  "1,2,x,y" ]`

To overcome such issues you need fewer loops and more Linq as readability will be greater so you can notice such issues easily:

```
        public IEnumerable<string> GetRecommendedServers(int schedulerId)
        {
            var schedulersToRunOnSameWebServer = HostController.Instance.GetString(
                "SchedulersToRunOnSameWebServer",
                string.Empty);

            var schedulers = schedulersToRunOnSameWebServer
                .Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)
                .Select(scheduler => scheduler.Trim());

            if (!schedulers.Contains(schedulerId.ToString()))
            {
                return Enumerable.Empty<string>();
            }
            else
            {
                var scheduleServerLists = schedulers
                    .Select(scheduler => int.TryParse(scheduler, out int id) ? id : (int?)null)
                    .Where(id => id != null && id != schedulerId)
                    .Select(id => SchedulingProvider.Instance().GetSchedule((int)id))
                    .Where(schedule => schedule != null && schedule.Enabled && !string.IsNullOrWhiteSpace(schedule.Servers));

                return scheduleServerLists
                    .SelectMany(schedule => schedule.Servers.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries))
                    // You are missing .Select(x => x.Trim()) before Distinct
                    .Distinct()
                    .OrderBy(x => x);
            }
        }
```

At least apply trim here or convert to like in the recommendation below,
With trim: `recommendedServers.AddRange(schedule.Servers.Split(new char[] { ',' }, StringSplitOptions.RemoveEmptyEntries)).Select(server => server.Trim())`

PR #1151, opened 2019-09-12
Delete style overflow after closing PersonaBar

### LINKED ISSUE #1030: PersonaBar closing action does not reset <body> overflow value properly
## Description of bug
When the PersonaBar is open the CSS style "overflow: hidden" is applied on the <body> tag. After closing the PersonBar the <body> tag gets the styling "overflow: auto;" which is IMHO a problem as "auto" is not the default value for overflow (it's "visible"). In other words: After opening and closing the PB, the body tag isn't in the same state as it was before. 

## Steps to reproduce
1. Inspect CSS overflow attribute on <body>
2. Open PersonaBar
3. Close PersonaBar
4. Inspect CSS overflow attribute on <body> again (it's now "auto" and not the value it was in step 1)

## Current result
Opening and closing the PersonaBar changes the CSS overflow attribute on <body>

## Expected result
After opening and closing the PersonBar, the CSS overflow attribute on <body> is in the state it was before opening the PersonaBar.

## Affected version
I only tested the following two versions, but most likely the problem exists in all versions that include the PersonaBar:

* [x] 9.4.0 RC1
* [x] 9.3.2

## Affected browser
The Problem is not browser specific.

tingung:
Fixes #1030 
## Summary

- Remove 'overflow' property instead change to 'auto'

Tested with all major browsers with latest version. (IE11, Google Chrome 76, Firefox 69, Edge 78)

Confirmation video - https://drive.google.com/file/d/1xZsGKRXLNfWQpuX5FTsM4Vihzmor6188/view


valadas (approved):
Looks good to me

valadas:
@tingung this is targetting development which means it would go for Dnn 10, would you like to resubmit this targetting release/3.0.x so it gets in Dnn 9.4.1 ?

tingung:
@valadas , let me check with stack holder and see if they would like to have it in 9.4.x. 

@daguiler , what do you think?

daguiler:
Sounds like a bugfix, so I would say 9.4.1

tingung:
Closing this PR - PR created for 3.0.x release target here - https://github.com/dnnsoftware/Dnn.AdminExperience/pull/1164

PR #1154, opened 2019-09-12
Bumps version to 9.4.1 for next release

valadas:
This is a release management task and we are self approving it as per our policies.


mitchelsellers (approved):
LGTM - Merging

PR #1156, opened 2019-09-12
Bumps version to 10.0.0 for next release

valadas:
This is a release management task and we are self approving it as per our policies.


valadas:
Should be handled by #1174 so closing this one

PR #1162, opened 2019-09-16
Modules get duplicated when try to localize modules

### LINKED ISSUE #1161: Modules get duplicated when try to localize modules
## Description of bug
Canon client mentioned that they're facing a problem when they try to Add missing language, as modules are getting duplicated as you can see in this [recording](https://drive.google.com/file/d/1fQIpsfGeQn33zDiCQxdl7Akz9o5RgEZ7/view)

## Steps to reproduce
1. Create a DNN page under the main language (en-us) 
2. Go to Page settings localization And Add missing languages 
3. Click all available Module checkboxes on both languages 
4. Click on "Update Localization" but do not wait till it finished, just refresh the page while it is processing 
5. The same page will be opened up and again check all available modules on both languages area 
6. Click the "Update Localization" button and this time wait till it gets finalized 
7. Check both pages

## Current result
Some modules are delivered to each other (Fr-FR modules moved to English or vice versa)

## Expected result
Both pages should have their relevant modules

## Screenshots
[recording](https://drive.google.com/file/d/1fQIpsfGeQn33zDiCQxdl7Akz9o5RgEZ7/view)

## Affected version

* [ ] 9.3.2
* [ ] 9.3.1
* [x] 9.2.2
* [x] 9.2.1
* [x] 9.2
* [x] 9.1.1
* [ ] 9.1
* [ ] 9.0

## Affected browser

* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

aelibyary:
Fixes #1161 

## Summary
Part if the issue happens mainly when the "Update Localization" operation takes so long and the user simply hit the refresh button thinking that the operation has failed or stuck,

@daguiler Said:
We will mitigate this issue with a short-term solution: we will warn the user that updating localization is a lengthy operation and they should refrain from refreshing the page while the process is still running. The wording could be:

Updating localization might take a few minutes depending on the number of modules. Please wait until the process finishes completely.

And there should be Cancel and Continue buttons.

Fix Confirmation Video: https://drive.google.com/file/d/1buCf2Khn3tIQHqSCFTUUXqq6BfHA-BOj/view


valadas (approved):
Looks good to me

donker (approved):
Improves it

PR #1167, opened 2019-09-19
Social Groups created through the Persona Bar do not appear in the Social Groups

SergeyDryomin:
Social Groups created through the Persona Bar do not appear in the Social Groups fixed
https://github.com/dnnsoftware/Dnn.Platform/issues/3007
https://jira.devfactory.com/browse/DNN-30220

## Summary
1) Create the role with Security Mode = Social Group, Status = Approved, Public
2) Get RoleID and UserId from created role.
3) Add  Groups folder, if it doesn't exists. Add the folder with name equal to RoleID. 
4) Add creator as owner to the Social Group(role)

### DIFF
--- a/Extensions/Content/Dnn.PersonaBar.Extensions/Components/Roles/RolesController.cs
+++ b/Extensions/Content/Dnn.PersonaBar.Extensions/Components/Roles/RolesController.cs
@@ -192,9 +192,10 @@ namespace Dnn.PersonaBar.Roles.Components
             Requires.NotNull("role", role);
 
             var parentFolder = FolderManager.Instance.GetFolder(role.PortalID, "Groups");
-            var portalFolder = FolderManager.Instance.GetFolder(role.PortalID, "");
+            
             if (parentFolder == null)
             {
+                var portalFolder = FolderManager.Instance.GetFolder(role.PortalID, string.Empty);
                 parentFolder = AssetManager.Instance.CreateFolder(role.RoleID.ToString(),
                     portalFolder.FolderID,
                     portalFolder.FolderMappingID,
@@ -204,7 +205,7 @@ namespace Dnn.PersonaBar.Roles.Components
             AssetManager.Instance.CreateFolder(role.RoleID.ToString(),
                 parentFolder.FolderID,
                 parentFolder.FolderMappingID,
-                "");
+                string.Empty);
 
             RoleController.Instance.AddUserRole(role.PortalID,
                 userId,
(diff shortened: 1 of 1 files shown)


Extensions/Content/Dnn.PersonaBar.Extensions/Components/Roles/RolesController.cs
```diff
@@ -170,9 +173,48 @@ public bool SaveRole(PortalSettings portalSettings, RoleDto roleDto, bool assign
...
+        /// Add Role to Social Group, create folder(s) for it and adds user created role as owner.
+        /// </summary>
+        /// <param name="role"></param>
+        /// <param name="userId"></param>
+        public void AddSocialRole(RoleInfo role, int userId)
+        {
+            Requires.NotNull("role", role);
+
+            var parentFolder = FolderManager.Instance.GetFolder(role.PortalID, "Groups");
+            var portalFolder = FolderManager.Instance.GetFolder(role.PortalID, "");
```
Yabgu: Use string.Empty instead of ""

SergeyDryomin: fixed

valadas:
I am just wondering if this is the right place for the fix. Social groups existed before the persona bar so I am guessing there is code in the core already for handling that. 

I am not a 100% sure since I do  not use social features much, but just trying to see if there would be a way to move this into the core. Was this bug pre-existing before the persona bar ? If so was this logic part of the UI too before ? Just trying to avoid duplication and breaking changes.

I believe the persona bar code should be more UI focused and everything else should be in the core to allow other extensions to benefit from the fix.

sleupold:
@valadas,
in DNN 6, Social Groups had been piggypacked onto Security Roles (although they are totally different objects), and I am not sure, whether some of the functionality might have been implemented in Roles admin module - and might not have been properly migrated to the PB module.
Anyways, if we ever touch Social Groups in DNN, we should move them to dedicated objects (database table) with references to member role, manager role etc.

SergeyDryomin:
Hi , @valadas @sleupold , implementation part can be moved to https://github.com/sergeydryomin/Dnn.Platform.git into DNN Platform\Library\Security\Roles\RoleController.cs

valadas:
@sergeydryomin will you be submitting a PR there ?

valadas (approved):
I looked a little more into this and the groups module does it about the same, so I guess there is nothing already existing in the platform that does all those steps. For that reason, I am fine with this PR.

mitchelsellers (approved):

PR #1170, opened 2019-09-24
Fix for upgrade issues of Personabar in 9.4.0

### LINKED ISSUE #1163: Upgrading to 9.4.0 does not always unregister the old PersonaBar extensions
## Description of bug
This relates to https://github.com/dnnsoftware/Dnn.Platform/issues/2993

When upgrading to 9.4.0 form 9.x.x the old Persona Bar modules do not always get unregistered. The code removes the dlls but the old modules stay registered.

## Steps to reproduce
1. Go to Extensions
2. Select "Persona Bar Modules" in the dropdown

## Current result
Dnn.PersonaBar.Extensions has version 3.0.0
All other modules are still listed with an older version like 2.0.1

## Expected result
Only Dnn.PersonaBarExtensions 3.0.0 should be registered since we merged all modules.

## Affected version

* [x] 9.4.0
* [ ] 9.3.1
* [ ] 9.2.2
* [ ] 9.2.1
* [ ] 9.2
* [ ] 9.1.1
* [ ] 9.1
* [ ] 9.0

donker:
Fixes #1163

### DIFF
--- a/Extensions/Content/Dnn.PersonaBar.Extensions/Dnn.PersonaBar.Extensions.csproj
+++ b/Extensions/Content/Dnn.PersonaBar.Extensions/Dnn.PersonaBar.Extensions.csproj
@@ -714,6 +714,9 @@
   <ItemGroup>
     <Content Include="SqlDataProvider\03.00.00.SqlDataProvider" />
   </ItemGroup>
+  <ItemGroup>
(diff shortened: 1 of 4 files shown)


valadas (approved):
Looks good to me but cannot test before it is merged.

david-poindexter (approved):
Look good - thanks @donker

PR #1173, opened 2019-09-26
Merging fixes from 9.4.1 into development

valadas:
This is a release management task and we are self-approving as per our policies


valadas:
Meh, too many conflicts will resubmit

Issue #1112, opened 2019-08-14
Persona Bar > Manage > Pages > Expand All Fails After Adding Multiple Pages


kkadir:
## Description of bug
The "Expand All" link stops functioning after adding page(s) through the Add Multiple Pages flow.

## Steps to reproduce
1. Log in as Super/Host user to Platform. 
2. Navigate to Persona Bar > Manage > Pages. 
3. Click on the Add Multiple Pages button. 
4. Input a new page name. 
5. Click the "Add New Page(s)" button. 
6. Observe that the new page(s) is/are added to the "Pages List" tree-view. 
7. Click on the "Expand All" link on top of the "Pages List" tree-view.

## Current result
1. All items in the "Pages List" tree-view are in the collapsed state 
2. Clicking on the "Expand All" link; 
   1. Does not change the link as "Collapse All" 
   2. Does not expand all child items in the "Pages List" tree-view

## Expected result
1. All items in the "Pages List" tree-view are in the collapsed state 
2. Clicking on the "Expand All" link should; 
   1. Change the link as "Collapse All" 
   2. Expand all child items in the "Pages List" tree-view

## Screenshots
N/A

## Error log
N/A

## Additional context
N/A

## Affected version
* [x] 9.4.0 RC

## Affected browser
* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

valadas:
Assigning this to the 10.0.0 milestone as per the PR, if things change we can update.

valadas:
Merged

PR #1113, opened 2019-08-14
DNN-28054: Fix Expand All Issue After Adding Multiple Pages

kkadir:
Fixes #1112 

## Summary
Introduce;
- `addPagesComplete` action,
- `FORCE_PAGE_LIST_LOAD` and `UNFORCE_PAGE_LIST_LOAD` action types,
- New cases in reducer regarding the  new action types,
- `multiplePagesAdded` state
To enable tracking the multiple pages adding flow.

Finally, assure that child lists for all items in the pages list are loaded by updating the logic slightly in the `toggleExpandAll()` function.

**[Confirmation Video](https://drive.google.com/open?id=1dyzkUhomc03vd0JtuPPihNL9xfyN_Tk1)**

PR #1175, opened 2019-09-26
Expand All Fails After Adding Multiple Pages

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.AdminExperience/issues/1112

Summary: 
Issue caused due to `isChildLoaded` state property that we miss to reset when new page is added via `Add multiple pages`. On adding pages, tree view comes with the new pages list, but `isChildLoaded` remains unchanged after the first load. On rendering, it assumes all child items are already loaded, so it does not trigger them to be re-created in the tree. It does not allow Collapse/Expand to work properly.

Solution is to check whether number of pages has been changed. If yes, we reset `isChildLoaded` to `false`. Depending on this flag application will re-create child items on rendering. 

Please, note, this is alternative and much simpler way instead of one proposed here https://github.com/dnnsoftware/Dnn.AdminExperience/pull/1113

See confirmation video: [DEMO](https://drive.google.com/file/d/1S8Jg95dboApz9CTHiUI9K53U9LleV8Jc/view?usp=sharing)


daguiler:
Oh this looks much better. WIl give it a try

daguiler (approved):
Works fine @mikebigun

bdukes (approved):

Issue #1122, opened 2019-08-20
Persona Bar > Manage > Pages > Expand/Collapse All Fails After Moving The Only Child Item Away


kkadir:
## Description of bug
"Expand/Collapse All" fails after moving the only child item away from a parent node.

## Steps to reproduce
List the steps to reproduce the behavior:
1. Log in as Super/Host user to Platform. 
2. Navigate to Persona Bar > Manage > Pages. 
3. Click on the Add Multiple Pages button. 
4. Input 1 Parent page and child page. 
5. Click the "Add New Page(s)" button. 
6. Observe that the new page(s) is/are added to the "Pages List" tree-view. 
7. Expand the newly added page from the "Pages List" tree-view. 
8. Drag the child page under any other page. 
9. Click on the "Collapse All" link on top of the "Pages List" tree-view.

## Current result
1. All items in the "Pages List" tree-view are collapsed 
2. The "Collapse All" link is not changed as "Expand All" 
3. Keeping clicking on the unchanged "Collapse All" link does not have any effect.

## Expected result
1. All items in the "Pages List" tree-view are collapsed 
2. The "Collapse All" is changed as "Expand All" 
3. Keeping clicking on the "Collapse All" or "Expand All" link is collapsing or expanding all "Pages List" tree-view items respectively.

## Screenshots
[Issue Reproduction Video](https://drive.google.com/open?id=1Y9kUwughc9zQTJTBee4CfVsAL1V3-uGj)

## Error log
N/A

## Additional context
The issue arises when the only child item is dragged away from any parent item. For instance, assume that we have a page list structure like the following;
```
Item-01
Item-02
    > Item-02-01
    > Item-02-02
Item-03
    > Item-03-01
```
When _Item-02-01_ is dragged and dropped under _Item-01_, there won't be any issues as _Item-02_ will still have _Item-02-02_ as a child. But when _Item-03-01_ is dragged and dropped to anywhere, the issue will be observed as the _Item-03_ parent item is now without a child item.

## Affected version
* [x] 9.4.0 RC

## Affected browser
* [x] Chrome
* [x] Firefox
* [x] Safari
* [x] Internet Explorer
* [x] Edge

valadas:
Merged

PR #1123, opened 2019-08-20
DNN-31960: Fix Collapse All Issue After Moving Away The Only Child

kkadir:
Fixes #1122 

## Summary
On expand/collapse toggling, assure that childListItems, isOpen and hasChildren attributes are removed for parent nodes without children.

[Confirmation Video](https://drive.google.com/open?id=1hGuGyPyXfH2A9kCqlT8scLkIeLRJrY8H)

## Note
This includes code-changes also from #1113 to keep them as separate PRs so that they can be merged separately without requiring any sequence.

PR #1176, opened 2019-09-26
Expand/Collapse All Fails After Moving Dragging Away The Only Child item

mikebigun:
fixes https://github.com/dnnsoftware/Dnn.AdminExperience/issues/1122

I isolated the fix from here https://github.com/dnnsoftware/Dnn.AdminExperience/pull/1123. It contains redundant changes that are not related to the issue itself.  
When child item has been moved to another item, its parent becomes not properly updated and keeps child properties unchanged. We need to reset few properties to defaults to make Collapse/Expand works properly. 

Confirmation video: [DEMO](https://drive.google.com/file/d/1wQQnH_pGpOWYu6Ij9C9AxaQJFPcx3Icc/view?usp=sharing)


daguiler:
This one is related to https://github.com/dnnsoftware/Dnn.AdminExperience/pull/1175

daguiler (approved):

mitchelsellers (approved):
Looks good, holding on merging until 9.4.1 release is complete

PR #1181, opened 2019-09-29
Re-merging changes from 9.4.1 into development

valadas:
This is a release management task and we are self-approving it for that reason.


valadas:
I will resubmit a better version

